PR #1, opened 2025-10-01
Migrate CI to Builkdite & Implement Code Signing

AliSoftware:
Closes [AINFRA-1320](https://linear.app/a8c/issue/AINFRA-1320/code-sign-wordpress-contributor-toolkit-electron-app-winmac)

## What

 - Migrate the CI setup from GitHub Actions to Buildkite CI, so that we can access our CI secrets and tooling for Electron code signing
 - Configure code signing for Windows build
 - Configure code signing + notarization for macOS build

## How

Detailed internal post: paaHJt-8Wn-p2 

 - Create `.buildkite/pipeline.yml` (+ `shared_pipeline_vars`) to replace the commands run by `.github/workflows/build.yml`, and delete the latter
 - Create a `.buildkite/commands/install_node_dependencies.sh` script to wrap `npm ci` around `restore_cache`/`save_cache`, to optimize the CI build time with cached node modules[^1]
 - Set up some Ruby/Node/Windows env config files
    - Create `.nvmrc` to specify the Nodejs version
    - Create the `.windows-10-sdk-version` to fix the version of the Windows SDK to use for compiling native node module dependencies
    - Create `.ruby-version` to fix the Ruby version used by the `fastlane` tooling
    - Create `.bundle/config` to specify the Ruby install path to `vendor/bundle`
    - Create a `Gemfile` to fix the `fastlane` installed version, then `bundle install` it.
    - Create the `Fastfile` to implement the lane calling `sync_code_signing`, which is taking care of installing the macOS certificate and provisioning profile from S3 + Dev Portal
 - Create a `.buildkite/commands/setup_windows_code_signing.ps1` PowerShell script that:
    - Installs the Windows SDK specified by the version file (to allow `node` and in particular `node-gyb` to compile native dependencies when needed)
    - Decode and install our `certificate.pfx` signing certificate from our AWS secret store
    - Set up the necessary `CSC_*` env vars for `electron-builder` to find that certificate and its password during Windows code-signing
 - Create a `.buildkite/commands/setup_macos_code_signing.sh` that
    - Install the ruby tools we need (namely `fastlane`)
    - Calls `fastlane`'s `sync_code_signing`
    - Export [the necessary env vars for `electron-builder` to be able to notarize the app](https://www.electron.build/mac#notarize)
 - Fix some issues with the `package.json`
    - Fix the Windows architecture used from `arm64` to `x86`
    - Add a `author` key to the package, to avoid `node` from complaining about it.

## Additional Manual Tasks

For this whole setup to work for Apple code signing, I also had to manually create the `org.wordpress.contributor.toolkit` App ID on the Apple Developer Portal, then use `fastlane`'s `sync_code_signing` (in `readonly:false` mode) to generate the necessary provisioning profile and code signing certificate and store it in our S3 secrets.

[^1]: Such a utility is [intended to be migrated to our `a8c-ci-toolkit` plugin at some point](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/96), but we still didn't have the time to consolidate that yet. That project could have been a good occasion to get back to it, but given the short deadline I guess that's not gonna be for today 😅 

## Testing

 - Open [the Buildkite build that ran on this PR](https://buildkite.com/automattic/experimental-wp-dev-env/builds/35/steps/canvas)
 - Click on the "macOS Build" job, then go to the "Artifacts" tab, and download the `dist/WordPress Contributor Toolkit-0.1.0-arm64-mac.zip` and `dist/WordPress Contributor Toolkit-0.1.0-arm64.dmg` files
 - Extract the `.zip` on your Mac, then double-click on the extracted `.app`, and verify that you don't have the macOS alert about "that app is damaged and can't be opened, you should delete it" blah blah.
    - (You will still have an alert about "This app is downloaded from the Internet, are you sure you want to open it", but that's ok and expected, just click OK on that one)
 - Open the `.dmg` on your Mac, then drag & drop the `.app` to your `Applications` folder, and finally double-click that app from the Finder to open it. Validate that you don't have the "This app is damaged" alert either.
 - Back in Buildkite, Click on the "Windows Build" job, then "Artifacts" tab, then download the `.exe` file from there
 - Open that `.exe` on a Windows machine (either a physical machine or a VM) to install the app, then launch the installed app.
    - Verify that Windows don't complain about the app not being code-signed
    - If you have the possibility, it'd actually be nice to test the app both on a ARM Windows machine and a x86 one, to validate the binary works on both
 - If you have the possibility (i.e. if you have a Linux machine or VM), do the same kind of test with the Linux artifact.


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+      - github_commit_status: { context: "macOS Build" }
+
+  ########################################################
+  # Linux build
+  ########################################################
+  - label: ":linux: Linux Build"
+    agents: { queue: default }
+    plugins: [$CI_TOOLKIT, $NVM_PLUGIN]
+    command: |
+      echo "~~~ Install dependencies"
```
iangmaia, line 62: ❓ Not for now I guess, but will we need to eventually also sign Linux binaries?

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+      .buildkite/commands/setup_windows_code_signing.ps1
+
+      echo "~~~ Install dependencies"
+      bash .buildkite/commands/install_node_dependencies.sh
+
+      echo "~~~ Build renderer"
+      npm run build:once
+
+      echo "~~~ Build Windows artifact"
+      npm run dist:win
```
iangmaia, line 22: 💭 Not necessarily for this PR, but maybe we can run a programmatic command in the build itself to validate the binary and make sure it is properly signed (a quick search showed something like `signtool verify` on Windows, for example).

.buildkite/pipeline.yml
AliSoftware: Good question. TBH I have no idea if that's a thing. I mean I'm sure signing binaries in Linux is a thing, but will any Linux distributions ever prevent you from running unsigned binaries, e.g. like macOS does?

And I don't think any of our other Electron products that may generate Linux binaries is signed anyway so far.

So I'm not sure if that will ever be necessary anytime I  the near future 🤷 So let's only worry about that only if and when it's gonna be a friction/problem for end users 😛

.buildkite/pipeline.yml
AliSoftware: Yeah I already ran those commands manually (see associated P2 post about how to install and run `signtool.exe` to validate the signature, with the `/pa` flag being key) to validate those builds…
…But I guess we could consider running those commands systematically in all job runs in CI for validation too indeed 🤔 It probably won't take long to run anyway …

.buildkite/pipeline.yml
iangmaia: > I mean I'm sure signing binaries in Linux is a thing, but will any Linux distributions ever prevent you from running unsigned binaries, e.g. like macOS does?

Yep, that's what I'm wondering, it indeed seems to be [distribution specific](https://stackoverflow.com/questions/75109878/do-i-need-to-digitally-sign-my-code-for-linux). If we ever get inquiries about it we can look into it 👍

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "electron-setup-wordpress-core",
+  "author": "Adam Zieliński",
```
iangmaia, line 3: Nice, it shows then on the Mac file info:

Screenshot 2025-10-02 at 12 33 32

iangmaia: On the subject of other app data, perhaps would be nice to setup in a future PR as it seems the generic electron icon is being used: https://www.electron.build/icons.html

iangmaia (approved):
Approving to unblock, the changes LGTM 👍 

Note: I have only checked the Mac binaries built by CI.

adamziel:
Thank you so much! I'll go ahead and merge to keep iterating on this. IF anything else needs polish, let's handle that in a separate PR

PR #6, opened 2025-11-08
GitHub PR workflow

adamziel:
Adds a UI path for connecting with GitHub and creating a PR:

Zrzut ekranu 2025-11-8 o 01 31 22

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
       },
       "devDependencies": {
         "concurrently": "^8.2.2",
-        "electron": "^31.3.0",
+        "electron": "^31.7.7",
         "electron-builder": "^24.13.3",
         "esbuild": "^0.23.0"
       }
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   },
   "devDependencies": {
     "concurrently": "^8.2.2",
-    "electron": "^31.3.0",
+    "electron": "^31.7.7",
     "electron-builder": "^24.13.3",
     "esbuild": "^0.23.0"
   }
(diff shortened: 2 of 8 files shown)


juanmaguitar:
For the record, since this has been open a while: the flow it proposed has landed in trunk, and its layout with it.

- #179 implemented the GitHub path — device-code sign-in, fork, commit and pull request.
- #207 rebuilt this pull request's layout on top of it: the diff takes the room and the destinations stand beside it. That shape came from here.

Two things this branch got right that took the new implementation a day of testing against production GitHub to rediscover: device flow is the right sign-in, and `repo` is the scope it needs — `public_repo` is documented to cover public-repository writes and is refused by ref creation.

Where the two differ:

| | this branch | trunk now |
| --- | --- | --- |
| Token | persisted in `electron-store` | main-process memory, forgotten on quit |
| The pull request | pushes a branch, then opens GitHub's compare form in the browser | created through the API; the app never leaves |

And one thing this branch handled that trunk does not: skipping `.github/workflows/` files, which need a scope beyond `repo`. Filed as #208 — with the note that skipping them silently trades a confusing failure for a pull request quietly missing part of the change, so trunk will likely name the case instead.

Leaving this open rather than closing it; that call belongs to a maintainer. Thanks for the groundwork — the layout in particular was right, and it is what the app ships now.

juanmaguitar:
Closing as superseded. The connect-to-GitHub-and-open-a-PR flow this proposed has since shipped on trunk, factored across dedicated modules (github-auth, github-http, github-pr, pr-files, and the renderer's github-account / pr-state / diff-highlight) rather than inlined into `main.js` and `index.jsx`. The branch is ~100 commits behind and can't be meaningfully rebased onto that. The multi-project generalisation of this same PR flow is now tracked in #251.

PR #7, opened 2025-11-08
First stab at supporting Gutenberg projects 

adamziel:
wip


juanmaguitar:
Closing as superseded. This branch predates the current architecture by ~100 commits — it targets the old single-project layout and the pre-modularisation `main.js`/`index.jsx`, so it isn't revivable as code. The intent behind it is sound, though, and is now captured fresh against today's trunk in #251. Thanks for the early sketch — it's what confirmed the idea was worth doing properly.

PR #22, opened 2026-07-03
Fix macOS code signing (CSSMERR_TP_NOT_TRUSTED)

mokagio:
See https://linear.app/a8c/issue/AINFRA-2597/fix-macos-code-signing-for-contributor-toolkit

I doubt code signing was always broken for macOS, so something must have changed (but where? on Apple's end? in the Electron tooling?) since we first implemented it.

Anyway, this PR implement a workaround that does the job. (Verified on my end from build https://buildkite.com/automattic/experimental-wp-dev-env/builds/72).

I also noticed the artifacts are humongous. Started an agent to investigate and open a PR for it. Likely some kind of dependency or build artifact that gets packaged when it should be ignored.

<details>
<summary>AI-generated details</summary>

## Rationale

The macOS build shipped **unsigned**: electron-builder skipped signing with
`cannot find valid "Developer ID Application" identity`, reporting the installed
cert as `CSSMERR_TP_NOT_TRUSTED`.

Root cause: the macOS 26 CI image ships no **"Developer ID Certification
Authority" (G2) intermediate**, and `match` installs only the leaf cert. With no
keychain in the search list able to build the chain `Developer ID App →
Developer ID CA G2 → Apple Root`, `security` can't trust the identity, so
electron-builder silently drops signing. Same failure — and same fix — as the
WordPress desktop (calypso) app.

## Fix

Fetch Apple's intermediate and `import_certificate` it into the temp keychain
`setup_ci` creates, gated to CI. A new `verify_code_signing` fastlane lane
(`codesign --verify` + Gatekeeper `spctl --assess` + Developer ID authority
assertion) runs after `npm run dist`, so an unsigned or mis-signed build fails
the job instead of shipping.

The temporary keychain/trust diagnostics step is still present for this run to
confirm the chain now resolves; it will be removed once CI is green.

## How to test

Watch the macOS build on the Buildkite run for this PR: the "Verify macOS code
signing" step must pass, and the "Code signing diagnostics" section should now
show `verify-cert` succeeding.

</details>


fastlane/Fastfile
```diff
@@ -18,6 +21,17 @@ BUILDS_FOLDER = File.join(PROJECT_ROOT_FOLDER, 'dist')
...
+  # macOS 26 CI images don't ship the "Developer ID Certification Authority" G2
+  # intermediate, and `match` installs only the leaf. Without the intermediate in
+  # a search-list keychain, `security` can't build the chain and evaluates the
+  # identity as CSSMERR_TP_NOT_TRUSTED — electron-builder then silently skips
+  # signing. Install it into the temp keychain `setup_ci` just created.
+  if FastlaneCore::Helper.is_ci?
+    cert_path = File.join(Dir.tmpdir, 'DeveloperIDG2CA.cer')
+    sh('curl', '-fsSL', 'https://www.apple.com/certificateauthority/DeveloperIDG2CA.cer', '-o', cert_path)
+    import_certificate(certificate_path: cert_path, keychain_name: 'fastlane_tmp_keychain')
+  end
```
mokagio: This fixed the issue [in WordPress Desktop](https://github.com/Automattic/wp-calypso/pull/110331). But it's rater odd and raises the quesion "how does it work in all other projects that do Mac code signing in CI?"

My AI did [quite a bit of experimentation at the time](https://github.com/Automattic/wp-calypso/pull/110244) to understand the issue, FWIW, but it definitely needs more looking into, as tracked by https://linear.app/a8c/issue/AINFRA-2511/investigate-and-ideally-remove-the-code-signing-keychain-access

fastlane/Fastfile
```diff
@@ -42,6 +56,32 @@ lane :setup_code_signing do |readonly: true|
...
+    # Gatekeeper accepts it — passes only when Developer ID signed *and* notarized+stapled.
+    sh('spctl', '--assess', '--type', 'execute', '--verbose=2', app)
+
+    # Signed by our identity specifically, not merely some other valid certificate.
+    details = `codesign --display --verbose=2 #{app.shellescape} 2>&1`
+    UI.user_error!("#{app} is not signed by '#{expected_authority}':\n#{details}") unless details.include?("Authority=#{expected_authority}")
+  end
+
+  UI.success("Verified code signing for #{apps.length} app bundle(s)")
+end
```
mokagio, line 72: We should extract this in a fastlane action as some point

fastlane/Fastfile
```diff
@@ -18,6 +21,17 @@ BUILDS_FOLDER = File.join(PROJECT_ROOT_FOLDER, 'dist')
...
+
+  # macOS 26 CI images don't ship the "Developer ID Certification Authority" G2
+  # intermediate, and `match` installs only the leaf. Without the intermediate in
+  # a search-list keychain, `security` can't build the chain and evaluates the
+  # identity as CSSMERR_TP_NOT_TRUSTED — electron-builder then silently skips
+  # signing. Install it into the temp keychain `setup_ci` just created.
+  if FastlaneCore::Helper.is_ci?
+    cert_path = File.join(Dir.tmpdir, 'DeveloperIDG2CA.cer')
+    sh('curl', '-fsSL', 'https://www.apple.com/certificateauthority/DeveloperIDG2CA.cer', '-o', cert_path)
+    import_certificate(certificate_path: cert_path, keychain_name: 'fastlane_tmp_keychain')
```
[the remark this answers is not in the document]
mokagio: Not a concern at this time for how our CI is implemented via transient dedicated VMs.

mokagio:
@AliSoftware @iangmaia I cannot add you as reviewers (maybe @juanmaguitar can?) but I'd appreciate a look. Thanks!

juanmaguitar:
> @AliSoftware @iangmaia I cannot add you as reviewers (maybe @juanmaguitar can?) but I'd appreciate a look. Thanks!

I have added @AliSoftware @iangmaia as contributors to the project and @iangmaia is now a reviewer of this PR

fastlane/Fastfile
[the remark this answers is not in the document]
AliSoftware: I'm also not convinced about the comment the AI wrote, because [a quick look into the fastlane codebase](https://github.com/fastlane/fastlane/blob/2f82435b315cabb0f63d134d10db30d0f1f76281/fastlane_core/lib/fastlane_core/cert_checker.rb#L42) makes me think that fastlane _does_ take care of downloading and installing the `DeveloperIDG2CA.cer` certificate, so this should not be necessary

Besides, there are better and safer ways download a file in Ruby than `sh('curl')`, and especially for the case of a Root CA certificate, we should definitively never download a CA from the internet without at least validating its checksum (since Root CAs are the root of the trust chain, so any MitM that might happen during download would have significant security injection impacts, like the ability to codesign a build as an official trusted a8c app using a hijacked certificate trust chain). And `fastlane match` already takes care of this sha check when it installs the Apple DevID CAs for us while this custom code reinventing the wheel doesn't.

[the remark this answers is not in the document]
mokagio: @AliSoftware Turns out the Root CA workaround was completely unnecessary... I started #30 to investigate and as the first step removed the workaround to show red in CI—but CI was green. On top of the code sign validation passing in CI, I also verified by downloading the artifacts.

My guess is that it was some weird confluence of macOS image and Ruby/Electron tooling. But I won't be investigating this further.

I removed the workaround and verified the builds in https://github.com/WordPress/experimental-wp-dev-env/pull/22/commits/ed78442a851b186fa95e5feac88acf9f36bc6b8f, available at https://buildkite.com/automattic/experimental-wp-dev-env/builds/93

fastlane/Fastfile
```diff
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
+require 'tmpdir'
```
iangmaia, line 3: It seems we don't need `tmpdir`? 🤔

fastlane/Fastfile
```diff
@@ -42,6 +45,32 @@ lane :setup_code_signing do |readonly: true|
...
+  UI.user_error!("No `.app` bundle found under #{BUILDS_FOLDER}/mac*/ to verify") if apps.empty?
+
+  apps.each do |app|
+    UI.message("Verifying code signature: #{app}")
+
+    # The signature is intact and satisfies its own designated requirement.
+    sh('codesign', '--verify', '--deep', '--strict', '--verbose=2', app)
+
+    # Gatekeeper accepts it — passes only when Developer ID signed *and* notarized+stapled.
+    sh('spctl', '--assess', '--type', 'execute', '--verbose=2', app)
```
iangmaia, line 64: Should this check guarantee that the notarization ticket is stapled here?

From what I could [find out](https://developer.apple.com/documentation/security/notarizing-macos-software-before-distribution), if we really want to check if the notarization ticket is stapled into the app, we should probably run something like `xcrun stapler validate`.

iangmaia (approved):
Overall looks good, left a [comment](https://github.com/WordPress/experimental-wp-dev-env/pull/22#discussion_r3586443617) about validating the staple that's perhaps worth looking into. Approving to unblock.

fastlane/Fastfile
mokagio: Thanks @iangmaia ! Tracked in https://linear.app/a8c/issue/AINFRA-2709 as part of the extraction to `release-toolkit`

fastlane/Fastfile
mokagio: Right! It was used in the cert hack, but obviously no longer needed.

@copilot open a new PR to remove this `require`

PR #25, opened 2026-07-06
Fix VS Code not detected in packaged app

### LINKED ISSUE #24: "Open directory in" doesn't detect VS Code in the packaged app
## Description

In the packaged (signed) build of the WordPress Contributor Toolkit, the "Open directory in" dropdown shows **VS Code (not installed)** even though VS Code is installed and normally used to open the project.

Image

## Steps to reproduce

1. Install the app from a signed build (e.g. from CI, not via `npm start`).
2. Open a site and click "Open directory in".
3. Observe that VS Code is listed as "not installed", while Cursor/PHPStorm (when installed) are detected correctly.

## Expected behavior

VS Code should be detected as installed and available in the dropdown, matching what happens when running the app via `npm start`.

## Actual behavior

VS Code shows as "not installed" only in the packaged app. Running the exact same project via `npm start` correctly detects VS Code.

## Screenshots

Packaged app (bug):
"Open directory in" → VS Code (not installed)

Dev mode via `npm start` (works correctly):
"Open directory in" → VS Code detected

juanmaguitar:
## Summary

Fixes #24 — VS Code was reported as "not installed" in the `Open directory in` dropdown when running the packaged/signed build, despite being installed and working correctly under `npm start`.

## Root cause

`checkEditorAvailable()` in `src/main.js` detects VS Code purely via `which code`, relying on `process.env.PATH` as inherited by the Electron main process:

- When run via `npm start` from a terminal, the process inherits the full shell PATH (populated by `.zshrc`/`.bash_profile`, nvm, Homebrew, etc.), where the `code` shim installed by VS Code's "Shell Command: Install 'code' command in PATH" typically lives (`/usr/local/bin`, `/opt/homebrew/bin`, or `~/.local/bin`).
- When launched from Finder/Dock (the packaged app's normal launch path), macOS gives the process a minimal default PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) that doesn't include any of those locations, so `which code` fails.
- Unlike PHPStorm and Cursor, VS Code had no fixed `/Applications/*.app` path fallback to fall back on.

## Fix

- Augment the `PATH` used for the `which` lookup on macOS with the common shim install locations (`/usr/local/bin`, `/opt/homebrew/bin`, `~/.local/bin`).
- Add the VS Code CLI binary bundled inside the app (`/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code`) as an additional fallback candidate, checked the same way PHPStorm/Cursor bundle paths are checked.
- Generalized `editorCommands` entries to accept either a single command/path or an array of candidates, tried in order.

Other platforms (Linux, Windows) and other editors (PHPStorm, Cursor) are unaffected — their detection logic is unchanged.

## Test plan

- [ ] On macOS, build/sign the app (or otherwise launch it outside a terminal so it gets the minimal default PATH) with VS Code installed via the standard `.app` + "Install 'code' command in PATH" flow, and confirm the "Open directory in" dropdown now detects VS Code.
- [ ] Confirm `npm start` still detects VS Code as before.
- [ ] Confirm PHPStorm and Cursor detection is unaffected.
- [ ] Confirm behavior on Linux/Windows is unchanged (command lookup unaffected by these changes).

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -505,7 +505,10 @@ ipcMain.handle('dir:open', async (_e, directoryPath) => {
 
 const editorCommands = {
 	vscode: {
-		darwin: 'code',
+		// 'code' is only resolved as a fixed app-bundle path is not checked; the
+		// CLI shim path is added as a fallback since apps launched from Finder/Dock
+		// don't inherit the shell PATH where VS Code's "code" shim usually lives.
+		darwin: ['code', '/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code'],
 		linux: 'code',
 		win32: 'code.cmd'
 	},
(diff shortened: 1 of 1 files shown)


juanmaguitar:
Superseded by #26 — decided to remove the "Open directory in" dropdown/editor-detection feature entirely for now instead of patching the VS Code detection.

PR #26, opened 2026-07-06
Remove "Open directory in" editor-detection dropdown

juanmaguitar:
## Summary

Removes the "Open directory in" dropdown (Finder/VS Code/PHPStorm/Cursor) and its underlying editor-availability detection. It was originally added in #5 alongside the unrelated "Submit patch" button, which stays untouched here.

## Why

The detection was unreliable in the packaged/signed build (#24 — VS Code wasn't detected there even though it worked under `npm start`, due to GUI-launched apps not inheriting the shell PATH). Rather than keep patching around PATH edge cases across platforms, this removes the feature for now since it isn't essential.

Supersedes #25, which patched the detection instead of removing it.

## Changes

- `src/main.js`: removed `dir:open`, `editor:check-available`, `editor:open` IPC handlers and the `editorCommands`/`checkEditorAvailable` helpers.
- `src/preload.js`: removed `openDirectory`, `openInEditor`, `checkEditorsAvailable` from the exposed API.
- `src/renderer/index.jsx` (+ rebuilt `index.js` bundle): removed the `availableEditors` state/effect, the "Open directory in" `DropdownMenu`, and its now-unused `chevronDown` icon import.

The "Submit patch" button/download-patch flow (also from #5) is unaffected — different IPC channels, different UI, no shared code.

## Test plan

- [ ] Confirm the site card no longer shows an "Open directory in" control.
- [ ] Confirm "Submit patch" still works as before.
- [ ] Confirm the app builds and runs via `npm start` with no console errors referencing removed APIs.


ashfame (approved):
I would have preferred to keep it, and expose the path right there, so that users who can't get it to open that way, can just copy over the path.

Now, it feels like its worse, everyone is forced to copy path even if "open in" link could have worked for them. I suggest re-thinking about it. But I respect your decision as the primary maintainer of the project :) so marking it as approved as far as review is concerned.

Whatever you decide, I think a follow up PR surfacing the path right there for easy copy would be a nice addition.

juanmaguitar:
> Now, it feels like its worse, everyone is forced to copy path even if "open in" link could have worked for them. I suggest re-thinking about it.

For context. This is not available in the latest released [build v0.1.0](https://github.com/WordPress/experimental-wp-dev-env/releases/tag/v0.1.0) so it would be a new feature (that doesn't work properly)

> I think a follow up PR surfacing the path right there for easy copy would be a nice addition.

I agree adding the path and a quick way to copy to the clipboard is a solid addition that doesn't add extra quirks. I opened https://github.com/WordPress/experimental-wp-dev-env/pull/29 for this

PR #29, opened 2026-07-07
Surface site path inline with a quick copy button

juanmaguitar:
## Summary

Follow-up to #26. As raised in [review](https://github.com/WordPress/experimental-wp-dev-env/pull/26#pullrequestreview-4639851048), removing the editor-detection dropdown means everyone now has to fall back to copying the site path manually — but that path was only reachable through the "More" menu. This surfaces it inline in the site header for quick access, with a one-click copy button (and "Copied!" feedback).

CleanShot 2026-07-07 at 10 49 34

## Changes

- `src/renderer/index.jsx` (+ rebuilt `index.js` bundle): added inline path display + copy button in the `SiteRow` header; the "More → Copy path" menu item now reuses the same `copyPath` handler instead of duplicating the clipboard logic.

## Test plan

- [ ] Confirm the site path is visible inline in the site header.
- [ ] Confirm clicking the copy button copies the path and shows "Copied!" briefly.
- [ ] Confirm "More → Copy path" still works.
- [ ] Confirm the app builds and runs via `npm start` with no console errors.

Base branch is `remove/editor-detection-dropdown` since this builds on top of it; GitHub will retarget to `trunk` automatically once that merges.

### DIFF
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -13912,7 +13912,7 @@
           }
           function pushStoreConsistencyCheck(fiber, getSnapshot, renderedSnapshot) {
             fiber.flags |= StoreConsistency;
-            var check = {
+            var check2 = {
               getSnapshot,
               value: renderedSnapshot
             };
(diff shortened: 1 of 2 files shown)


ashfame (approved):
LGTM! The copy button didn't give feedback if actual copy to clipboard action has happened though. So, I had to paste to confirm. A little animation or changing of icons for 2secs will do the trick.

juanmaguitar:
> . A little animation or changing of icons for 2secs will do the trick.

Done ✅

Electron

PR #32, opened 2026-07-22
Added a persistent "Share feedback" button linking to a Google Form

### LINKED ISSUE #31: Add a persistent "Share feedback" button linking to a Google Form
Right now the only way we learn whether this toolkit is actually helping is by asking table leads after a Contributor Day — slow and indirect. There's no in-app feedback mechanism at all today.

**Proposal:** add a small, always-visible "💬 Share feedback" button (outlined style) somewhere persistent in the UI — e.g. the sidebar header, alongside the existing collapse toggle. Clicking it opens a lightweight popover (not a modal) with:
- A short note on why feedback matters ("Your feedback helps decide what to build next.")
- What happens to responses ("Responses go into a shared form the team reviews regularly. Submissions are anonymous unless you add your email.")
- A link: "Open the feedback form ↗"

Image

Image

**Why a persistent button instead of a one-time prompt:** it stays available on demand whenever someone has something to say, rather than interrupting once and being gone. It also needs no "have we shown this already" state to track.

**Implementation notes:**
- The form link opens via the app's existing `window.api.openExternal()` path (already used for Playground/dev-server URLs) — no new IPC needed.
- The popover can be a lighter-weight component than the app's existing `@wordpress/components` `Modal` dialogs (anchored to the button, dismiss on outside click) — no new heavy UI dependency either way.
- Exact placement is left to whoever implements this; sidebar header is one option.

juanmaguitar:
Fixes #31 

Adds a 💬 "Share feedback" link to the site row UI that opens a Google Form in the user's default browser via `window.api.openExternal`. 

The form URL is https://docs.google.com/forms/d/e/1FAIpQLScnMxicyDxZO2OoaS5ela8FArYWjCyLfC3hxRBBRSF7XLPzKg/viewform

CleanShot 2026-07-22 at 11 28 12

The form URL is exposed to the renderer through a new `config:get-feedback-form-url` IPC handler in the main process (`src/main.js`) and bridged in `src/preload.js`; the URL is hardcoded rather than sourced from `.env`/`process.env`, since it's a static, non-secret value and hardcoding avoids `dotenv`'s unreliable `cwd`-relative resolution in packaged builds.

- `src/main.js` — new IPC handler returning the feedback form URL
- `src/preload.js` — exposes `getFeedbackFormUrl()` to the renderer
- `src/renderer/index.jsx` — fetches the URL on mount, renders the link in `SiteRow`, disabled/greyed out if unset
- `package.json` / `package-lock.json` — `dotenv` dependency removed (no longer used anywhere in the app)

### Test plan

- [ ] `npm start`, confirm "Share feedback" link appears on a site row and opens the Google Form in the default browser

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4980,16 +4980,6 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/dotenv": {
-      "version": "9.0.2",
-      "resolved": "https://registry.npmjs.org/dotenv/-/dotenv-9.0.2.tgz",
-      "integrity": "sha512-I9OvvrHp4pIARv4+x9iuewrWycX6CcZtoAu1XrzPxc5UygMJXJZYmBsynku8IkrJwgypE5DGNjDPmPRhDCptUg==",
-      "dev": true,
-      "license": "BSD-2-Clause",
-      "engines": {
-        "node": ">=10"
-      }
-    },
     "node_modules/dotenv-expand": {
       "version": "5.1.0",
       "resolved": "https://registry.npmjs.org/dotenv-expand/-/dotenv-expand-5.1.0.tgz",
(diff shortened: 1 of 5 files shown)


ashfame (commented):
Thanks for adding the feedback entry point. I found two issues that should be addressed before this closes #31:

- **[P1] Keep the feedback control outside the `skipInit`/site-row gate.** It is currently rendered only inside the post-wizard site controls, so it is absent when there are no sites and throughout the initial setup flow. That conflicts with the issue requirement for an always-visible, persistent control and hides it from first-time users. Please move it to an application-level location such as the sidebar header.

- **[P2] Show the requested explanatory popover before opening Google Forms.** The current click handler calls `openExternal` immediately, omitting the required context about why feedback matters, how responses are reviewed, and anonymity. The button should open the lightweight popover, and the link inside that popover should invoke `openExternal`.

Validation otherwise passed: clean dependency install, all 13 tests, syntax checks, `git diff --check`, and a reproducible renderer bundle.

juanmaguitar:
Closed in favour of #33

PR #38, opened 2026-07-27
Retry npm install with relaxed engine checks when a dependency demands a newer Node

### LINKED ISSUE #8: Sites fail to start dev server
When trying the app, I followed the steps to install and build, but the site server doesn't start when I click `Start dev server`. Here are the logs:

```
❌ npm ci failed: npm ci failed with code 1
❌ Sync failed: node tools/gutenberg/checkout-gutenberg.js failed with code 1
Warning: Task "gutenberg-sync" failed. Use --force to continue.
```

Do you know if I'm doing anything wrong? I'm using trunk in MacOS arm.

Image

### LINKED ISSUE #37: "Install npm dependencies" fails with EBADENGINE: bundled Node is older than wordpress-develop requires
## Summary

"Install npm dependencies" fails for every user with `EBADENGINE`, leaving new sites unusable. Reproduced on `trunk` and on the shipped [v0.1.0 release](https://github.com/WordPress/experimental-wp-dev-env/releases/tag/v0.1.0).

```
npm error code EBADENGINE
npm error engine Unsupported engine
npm error notsup Not compatible with your version of node/npm: @eslint/compat@2.1.0
npm error notsup Actual:   {"npm":"10.9.3","node":"v20.18.0"}
```

## Findings

Nothing in this app regressed — `wordpress-develop`'s dependency tree moved under a Node version we pin.

Two conditions combine:

1. Child `npm install` runs on Electron's bundled Node (`process.execPath` + `ELECTRON_RUN_AS_NODE=1`). Electron `^32.3.3` bundles **Node v20.18.1** (v20.18.0 in the v0.1.0 build) — short of the required `20.19.0` by two patch versions.
2. `wordpress-develop`'s `.npmrc` sets **`engine-strict = true`**, so an `engines` mismatch is a fatal error rather than a warning.

The requirement arrived upstream:

| Date | Event |
|---|---|
| 2026-01-16 | `@wordpress/eslint-plugin@24.0.0` — no `@eslint/compat` dependency. App worked. |
| 2026-04-15 | `@wordpress/eslint-plugin@25.0.0` adds `@eslint/compat: ^2.0.0`, requiring Node `^20.19.0 \|\| ^22.13.0 \|\| >=24`. **Breaks here.** |
| 2026-07-14 | `wordpress-develop` now locks `@wordpress/eslint-plugin@25.7.0` → `@eslint/compat@2.1.0`. |

Dependency chain: `wordpress-develop` → `@wordpress/eslint-plugin` → `@eslint/compat`. Nothing declares it directly, and `wordpress-develop`'s own `package.json` declares only `"node": ">=20.10.0"`, so the real floor is invisible from its declared engines.

## Notes

- `@eslint/compat` is lint-only — not required to build WordPress, run the dev server, or produce a patch.
- Patch generation is unaffected; it only needs the clone.
- Because the app pins a Node version while tracking a moving `trunk`, this is a recurring class of failure rather than a one-off.
- Possibly the same root cause as #8 (`npm ci failed` inside the Gutenberg sync grunt task) — unverified.

juanmaguitar:
Fixes #37.

## Problem

"Install npm dependencies" fails for everyone with `EBADENGINE` before installing anything, on `trunk` and on the shipped v0.1.0. Nothing here regressed — `wordpress-develop`'s dependency floor rose while the Node version Electron bundles stayed put:

- Child `npm install` runs on Electron's bundled Node, which is **v20.18.1**.
- `@wordpress/eslint-plugin@25.0.0` (2026-04-15) pulled in `@eslint/compat@^2.0.0`, which requires Node `^20.19.0 || ^22.13.0 || >=24`.
- `wordpress-develop`'s `.npmrc` sets `engine-strict = true`, making the mismatch fatal rather than a warning.

Upgrading Electron raises the floor but doesn't remove the failure mode — it just resets the clock until the next upstream bump. This makes the app tolerant of the drift instead.

## Approach

Run the install normally. **Only** if it fails for an engines mismatch, retry once with `npm_config_engine_strict=false` and say so in the log:

```
npm error code EBADENGINE

⚠ This site requires a newer Node.js than this app bundles.
  Retrying with engine checks relaxed…

added 2566 packages, and audited 2567 packages in 14s
```

The first failure still reaches the log, so the real reason stays visible rather than being permanently silenced. Relaxing the check is low-risk here: the packages raising the floor are all lint-only, and aren't needed to build WordPress, run the dev server, or produce a patch.

Two decisions worth a reviewer's attention:

- **`npm_config_engine_strict` is set in the child's environment, not passed on npm's command line.** Env vars are inherited, so it also reaches nested `npm` calls made by WordPress's own scripts; a flag would not. It may therefore fix #8 as well (unverified).
- **`npm:kill` records the cancellation before terminating the child.** Stopping a script exits non-zero with the engines warning already logged — which a naive retry would otherwise take as a reason to start the work over.

## Changes

- New `src/npm-runner.js` — spawn env and retry decision, free of Electron imports so it's unit-testable (same seam as `src/patch.js`). Also de-duplicates the two identical env blocks `npm:install` and `npm:run-script` each carried.
- `src/main.js` — `runNpmWithEngineRetry()` wraps both handlers, reusing the same `installId`/`runId` so the renderer's log stream stays continuous.
- Tests: 25/25 pass (13 existing + 12 new).

## Test plan

- [x] `npm test` — 25/25 pass
- [x] Verified end-to-end against a real `wordpress-develop` clone with an empty `node_modules`: first attempt fails, retry fires, **install completes** with 2566 packages and exit 0.
- [x] `node --test test/npm-install.integration.test.cjs` — runs the real install runner against a synthetic fixture reproducing the same `EBADENGINE`, in ~200ms with no network.

**Not covered:** verified on macOS only — `ensureNodeShimDir()`'s Windows shims stay untested, and `npm test` runs in no Buildkite step yet (#39).

### DIFF
--- a/src/npm-runner.js
+++ b/src/npm-runner.js
@@ -66,17 +66,21 @@ function buildChildEnv({
 } = {}) {
 	const isWindows = platform === 'win32';
 	const separator = isWindows ? ';' : ':';
+	const basePath = isWindows
+		? (baseEnv.Path || baseEnv.PATH || '')
+		: (baseEnv.PATH || '');
+	const joinedPath = basePath ? `${shimDir}${separator}${basePath}` : String(shimDir);
 	const env = {
 		...baseEnv,
 		ELECTRON_RUN_AS_NODE: '1',
 		NODE: execPath,
 		npm_config_production: 'false',
 		NODE_ENV: 'development',
-		PATH: `${shimDir}${separator}${baseEnv.PATH || ''}`,
+		PATH: joinedPath,
 		PATHEXT: isWindows ? WINDOWS_PATHEXT : baseEnv.PATHEXT
 	};
 	if (isWindows) {
-		env.Path = `${shimDir};${baseEnv.Path || baseEnv.PATH || ''}`;
+		env.Path = joinedPath;
 	}
 	return { ...env, ...extraEnv };
 }
--- a/src/npm-runner.js
+++ b/src/npm-runner.js
@@ -31,8 +31,11 @@ function createEngineMismatchDetector() {
 	let found = false;
 	return {
 		push(text) {
-			if (!found && isEngineMismatch(tail + text)) found = true;
-			tail = String(text || '').slice(-CHUNK_OVERLAP);
+			// The retained tail must be a suffix of the combined text, not of the
+			// newest chunk alone, or a marker spread over three+ chunks is lost.
+			const combined = tail + String(text || '');
+			if (!found && isEngineMismatch(combined)) found = true;
+			tail = combined.slice(-CHUNK_OVERLAP);
 			return found;
 		},
 		get found() {
--- a/src/npm-runner.js
+++ b/src/npm-runner.js
@@ -1,0 +1,90 @@
+// Helpers for spawning npm on Electron's bundled Node runtime.
+// Deliberately free of Electron imports so it can be unit-tested.
+
+const WINDOWS_PATHEXT = [
+	'.COM', '.EXE', '.BAT', '.CMD', '.VBS', '.VBE', '.JS', '.JSE', '.WSF', '.WSH', '.MSC'
+].join(';');
+
+// Relaxes `engine-strict`, which wordpress-develop's .npmrc turns on. Passed as
+// an environment variable rather than a CLI flag so it also reaches nested npm
+// invocations made by WordPress's own scripts (e.g. the Gutenberg sync npm ci).
+const RELAXED_ENGINES_ENV = { npm_config_engine_strict: 'false' };
+
+// True when npm output blames an engines mismatch. Kept narrow on purpose: a
+// failure for any other reason must not trigger the relaxed-engines retry.
+// Note npm prints EBADENGINE as a warning too when engine-strict is off, so
+// callers must also require a non-zero exit code.
+function isEngineMismatch(text) {
+	if (!text) return false;
+	return /EBADENGINE/.test(text)
+		|| /Not compatible with your version of node/i.test(text);
+}
+
+// Longest phrase isEngineMismatch looks for is ~40 chars; 64 is a safe overlap.
+const CHUNK_OVERLAP = 64;
+
+// Stateful detector for streamed output. Child stdout/stderr arrives in
+// arbitrary chunks, so a marker can straddle a chunk boundary; carrying a short
+// tail across calls keeps it detectable without buffering the whole log.
+function createEngineMismatchDetector() {
+	let tail = '';
+	let found = false;
+	return {
+		push(text) {
+			if (!found && isEngineMismatch(tail + text)) found = true;
+			tail = String(text || '').slice(-CHUNK_OVERLAP);
+			return found;
+		},
+		get found() {
+			return found;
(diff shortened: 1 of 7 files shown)


src/npm-runner.js
```diff
@@ -0,0 +1,94 @@
...
+// Stateful detector for streamed output. Child stdout/stderr arrives in
+// arbitrary chunks, so a marker can straddle a chunk boundary; carrying a short
+// tail across calls keeps it detectable without buffering the whole log.
+function createEngineMismatchDetector() {
+	let tail = '';
+	let found = false;
+	return {
+		push(text) {
+			if (!found && isEngineMismatch(tail + text)) found = true;
+			tail = String(text || '').slice(-CHUNK_OVERLAP);
```
ashfame: The retained tail currently comes only from the newest chunk, so a marker split across three chunks is lost. For example, `push("EBA")`, `push("DEN")`, `push("GINE")` leaves `found === false`. Since the helper is intended to handle arbitrary stream chunking, build a combined value first, test it, and retain its suffix: `const combined = tail + String(text || ""); ... tail = combined.slice(-CHUNK_OVERLAP);`. Please add a regression test for a three-chunk split.

ashfame (changes requested):
Thanks for the thorough explanation and test coverage. The environment-variable approach is sensible, cancellation is handled carefully, and the latest commit addresses the PATH/Windows feedback. I verified the current head with `npm test` (25/25 passing) and `git diff --check`.

I’m requesting changes for the streamed detector issue noted inline. It is small to fix, but it can prevent the recovery path from activating even though this helper explicitly accounts for arbitrary stream chunking.

One broader design concern to consider: applying the retry to `npm:run-script` can restart a partially executed `build`, `grunt`, or other script, and warning-form `EBADENGINE` output followed by an unrelated non-zero exit can trigger that retry. I would prefer limiting the automatic fallback to installation or requiring evidence of a fatal `EBADENGINE` for script runs. If the broader behavior is intentional, please document and test that tradeoff explicitly.

src/npm-runner.js
juanmaguitar: You're right — the tail came from the newest chunk instead of the combined text, so a marker spread over three chunks was lost. Fixed as suggested in https://github.com/WordPress/experimental-wp-dev-env/pull/38/commits/693312588c16b49870a6254f08910880b1730852, with the three-chunk regression test (EBA / DEN / GINE), which I confirmed failed before the fix.

juanmaguitar:
@ashfame as per https://github.com/WordPress/experimental-wp-dev-env/pull/38#pullrequestreview-4796035895

Both points addressed in 6933125.

**Detector:** fixed as you suggested, plus the three-chunk regression test.

**Retry scope:** agreed, now limited to installs. `runNpmWithEngineRetry` takes a `retryOnEngineMismatch` flag (default `false`) and only `npm:install` opts in. With the flag off the detector isn't fed at all, so `npm:run-script` behaves exactly as it did before this PR.

`npm test` green at 26/26 locally.

ashfame (approved):

PR #45, opened 2026-07-28
Harden npm engine retry behavior

ashfame:
## Summary

- Preserve the streamed engine-mismatch detector’s rolling tail across any number of chunks.
- Keep relaxed-engine retries opt-in for dependency installation so `npm run` commands are never automatically restarted after partial execution.
- Add regression coverage for a marker split across three chunks and for callers disabling retries.

## Why

This follows up on the review of #38. The detector previously retained only the newest chunk, so an `EBADENGINE` marker split across three chunks could be missed. The shared retry wrapper also applied to arbitrary npm scripts, which could restart a partially completed build, test, or Grunt task and could misclassify warning-form `EBADENGINE` output followed by an unrelated failure.

This PR is intentionally stacked on #38 and targets `juanmaguitar/issue-37-npm-engine-retry`; merging it updates PR 38 directly.

## Validation

- `npm test` — 27/27 passing
- `node --check src/main.js`
- `node --check src/npm-runner.js`
- `git diff --check`

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -513,11 +513,10 @@ ipcMain.handle('url:open', async (_e, url) => {
 
 const ENGINE_RETRY_NOTICE = '\n⚠ This site requires a newer Node.js than this app bundles.\n  Retrying with engine checks relaxed…\n\n';
 
-// Spawns an npm runner, and if it fails specifically because a dependency
-// demands a newer Node than Electron bundles, retries once with engine checks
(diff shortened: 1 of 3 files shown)


ashfame:
Closing as largely this has been included in #38 via https://github.com/WordPress/experimental-wp-dev-env/pull/38/commits/693312588c16b49870a6254f08910880b1730852

PR #50, opened 2026-07-29
Write an app log to disk and add a Help menu to open it

### LINKED ISSUE #49: Add a menu entry to open the app log (and write one to disk)
## What problem does this address?

There is currently no way to see what the app is actually doing when something goes wrong, which makes bugs very hard to diagnose — especially on Windows, and especially when helping a contributor remotely.

Concretely:

- The app doesn't define its own `Menu` (`src/main.js` never imports `Menu`), so we get Electron's default menu. DevTools are reachable via `View → Toggle Developer Tools`, but that only shows the **renderer** console.
- Everything interesting for the current Windows bugs happens in the **main** process: the `console.log` calls in `src/main.js:642-662`, the `spawn` calls for `npm install` / `npm run <script>` / the Playground server, and their stdout/stderr. In a packaged GUI `.exe` that output goes nowhere and is never written to disk.
- Child-process output is only forwarded to the renderer over IPC (`npm:install:log`, `npm:run-script:log`, `playground:log`), so if a spawn fails before producing output — or if the renderer never renders it, as in #48 — there is no record at all.

The practical consequence: "nothing happened" and "the spawn failed with EPERM" look identical to the user, and there is nothing to attach to a bug report. At a Contributor Day that turns into a dead end for the person helping.

## What is your proposed solution?

1. **Log to a file.** Write to `app.getPath('logs')` (either `electron-log` or a small `fs.createWriteStream` helper). It should capture the main process's `console.*` output **and** the stdout/stderr of every child process (install, run-script, Playground server, SMTP), which today are only sent over IPC to the renderer.
2. **Expose it in the menu.** Define an app `Menu` (Electron's default template plus a **Help** submenu) with:
   - `Open log file` → `shell.openPath(...)`
   - `Show logs folder` → `shell.showItemInFolder(...)`

   `shell` is already imported in `src/main.js:1` and used by the `url:open` handler (`src/main.js:500`). This works identically on macOS and Windows.
3. **Nice to have:** an explicit `Help → Toggle Developer Tools` entry, so a contributor who doesn't know the shortcut can find it.

This makes remote debugging a matter of "send me the log file" instead of "describe what you saw", and it's a prerequisite for diagnosing issues like #48 where the failure is invisible by construction.

juanmaguitar:
> **Stack — 1 of 5** · `trunk` → **#50** → #51 → #52 → #56 → #58

Closes #49

## Problem

A packaged build has no terminal, so nothing the app does is recorded anywhere. Child-process output is only forwarded to the renderer over IPC — it exists as pixels in the UI, and only if the UI manages to draw it.

So a spawn that fails with `EPERM` and a spawn that never started look identical to the person in front of the app: nothing happens. Helping someone remotely becomes "describe what you saw", and at a Contributor Day that is where the session ends.

## Fix

The app writes a log file, and **Help → Open App Log** takes you to it. Everything that was previously invisible now lands there: child-process output, the command used to spawn each one and the code it exited with, renderer errors, and crashes with stack traces.

Asking for a bug report becomes "open Help, click Show Logs Folder, attach that file".

[CleanShot 2026-07-29 at 11 14 25@2x
](https://www.youtube.com/watch?v=mJGF8S3cLhQ)
[see video](https://www.youtube.com/watch?v=mJGF8S3cLhQ)

## How to test

Run the app from this branch: `npm install && npm start`

1. Open **Help**. It has three entries: *Open App Log*, *Show Logs Folder*, *Toggle Developer Tools*.
2. Click **Open App Log**. The file opens in your default viewer and starts with a session header — app version, Electron, Node, Chrome, platform, arch.
3. Click **Show Logs Folder**. The file is revealed in Finder/Explorer, named `wordpress-contributor-toolkit-<platform>.log`.
4. Create a site and click **Install npm dependencies**. While it runs, the log fills with the same output you see in the Terminal panel, plus two lines the UI never shows: `spawn install-runner.js <path>` when it starts and `exited with code 0` when it ends.
5. Quit and reopen the app. A new `--- session start ---` header is appended; the previous session is still there.

Automated: `npm test` (13 new cases).

<details>
<summary><b>The menu, the file location, and what ends up in it</b></summary>

Under **Help**:

- **Open App Log** — opens the log in the system's default viewer
- **Show Logs Folder** — reveals it in Explorer/Finder, for attaching to an issue
- **Toggle Developer Tools** — the same item as under View, where someone who doesn't know the shortcut will actually look for it

The file is named `wordpress-contributor-toolkit-<platform>.log` so it still identifies itself once it has been detached from its folder and attached to a GitHub issue alongside logs from other contributors.

| Platform | Path |
| --- | --- |
| Windows | `%APPDATA%\WordPress Contributor Toolkit\logs\wordpress-contributor-toolkit-windows.log` |
| macOS | `~/Library/Logs/WordPress Contributor Toolkit/wordpress-contributor-toolkit-macos.log` |
| Linux | `~/.config/WordPress Contributor Toolkit/logs/wordpress-contributor-toolkit-linux.log` |

What ends up in it:

- every session starts with app, Electron, Node, Chrome, platform and architecture — the block every bug report needs and nobody remembers to include
- the full output of `npm install`, `npm run <script>` and both Playground servers, as whole lines rather than the fragments the streams actually arrive in
- the command used to spawn each child process and the code it exited with, so a child that dies before printing anything is still visible — that silence being precisely the bug on Windows today
- errors from the UI as well as the main process, so a failure where the app is working fine and the interface simply never shows it is no longer invisible
- crashes: uncaught exceptions and unhandled rejections in both processes, with stack traces, where previously a crash left an empty log

It rotates at 5 MB, so a long-lived session cannot fill someone's disk. It writes alongside the existing UI output rather than replacing it — the terminal panel behaves exactly as before.

</details>

<details>
<summary><b>Notes for reviewers</b></summary>

**The log is not redacted.** It contains local paths (`C:\Users\<name>\…`) and environment details, in a file users are directed to attach to public issues. That is the same information a bug report needs, so it seems the right trade, but it is a deliberate choice rather than an oversight.

**"App Log" is deliberately not "Log".** The app already tails each site's WordPress `debug.log` and shows it in the UI; these are different files for different purposes, and the wording is there to keep people from opening the wrong one.

**Setting an application menu replaces Electron's default wholesale.** The default submenus are rebuilt from built-in roles rather than dropped, so nothing that worked before disappears — worth a look on both platforms, since a mistake there presents as a missing or empty menu.

**Anyone on an earlier build of this branch has an orphaned `main.log`** in that folder from before the file was renamed. Nothing cleans it up, so two files in the folder is expected rather than a bug.

</details>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@wp-playground/cli": "^3.0.8",
         "@xterm/xterm": "^5.5.0",
         "diff": "^5.2.0",
+        "electron-log": "^5.4.4",
         "electron-store": "^10.0.1",
         "extract-zip": "^2.0.1",
         "fs-extra": "^11.2.0",
@@ -5138,6 +5139,15 @@
       },
       "engines": {
         "node": ">=12"
+      }
+    },
+    "node_modules/electron-log": {
+      "version": "5.4.4",
+      "resolved": "https://registry.npmjs.org/electron-log/-/electron-log-5.4.4.tgz",
+      "integrity": "sha512-istWgaXjBfURBSS8LWVW9C3jsc6+ac+tY1lXrQEOTp0lVj+a4OlO1Tmqb36GgnEUDv92DGC9VI1HNXwJinWpgA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">= 14"
       }
     },
     "node_modules/electron-publish": {
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@wp-playground/cli": "^3.0.8",
     "@xterm/xterm": "^5.5.0",
     "diff": "^5.2.0",
+    "electron-log": "^5.4.4",
     "electron-store": "^10.0.1",
     "extract-zip": "^2.0.1",
     "fs-extra": "^11.2.0",
--- a/src/log-lines.js
+++ b/src/log-lines.js
@@ -1,0 +1,34 @@
+// Reassembles child-process stream chunks into whole lines for the app log.
+// Deliberately free of Electron imports so it can be unit-tested.
+
+// Child stdout/stderr arrives in arbitrary chunks: a single npm message can be
+// split across two reads, and one read can carry a dozen messages. Logging the
+// chunks verbatim would stamp a timestamp mid-sentence and interleave the two
+// streams mid-word, which is exactly what makes a log unreadable at the moment
+// someone needs it. Buffering to line boundaries costs one partial line of
+// latency and is why `flush` exists — the last line before a process exits
+// usually has no trailing newline, and it is often the interesting one.
+function createLineBuffer() {
+	let pending = '';
+	return {
+		// Returns the complete lines contained in `chunk`, retaining any partial
+		// trailing line for the next call.
+		push(chunk) {
+			pending += String(chunk ?? '');
+			// A trailing \n yields a final '' element, which is dropped as the new
+			// pending — so a chunk ending exactly on a boundary leaves nothing behind.
+			const parts = pending.split('\n');
+			pending = parts.pop();
+			// \r\n from Windows child processes would otherwise show as a stray CR.
+			return parts.map((line) => line.replace(/\r$/, ''));
+		},
+		// The unterminated remainder, if any. Call once when the stream closes.
+		flush() {
+			const rest = pending.replace(/\r$/, '');
+			pending = '';
+			return rest ? [rest] : [];
+		}
+	};
+}
+
+module.exports = { createLineBuffer };
--- a/src/logging.js
+++ b/src/logging.js
@@ -1,0 +1,127 @@
+// The app's on-disk log. This is the only module that touches electron-log, so
+// the rest of the codebase depends on these four functions rather than on the
+// library.
+//
+// Why a file at all: in a packaged GUI build the main process has no terminal,
+// so every console.* call and every byte of child-process stdout/stderr goes
+// nowhere. Child output is forwarded to the renderer over IPC, which means a
+// spawn that fails before producing output — or a renderer that never renders
+// what it received — leaves no trace whatsoever. "Nothing happened" and "the
+// spawn failed with EPERM" then look identical, and there is nothing to attach
+// to a bug report.
+
+const path = require('path');
+const { app } = require('electron');
+const log = require('electron-log/main');
+const { createLineBuffer } = require('./log-lines');
+
+// Rotated to main.old.log past this size, so an app left running through many
+// npm installs cannot grow the file without bound.
+const MAX_LOG_SIZE = 5 * 1024 * 1024;
+
+let initialized = false;
+
+function initLogging() {
+	if (initialized) return;
+	initialized = true;
+
+	// Must run before the first BrowserWindow is constructed: this preloads the
+	// IPC bridge that lets renderer output reach this file. Called later it
+	// silently does nothing for windows that already exist. spyRendererConsole
+	// is what captures the renderer's own console.* — the blind spot behind the
+	// class of bug where the main process is fine and the UI simply never shows
+	// what it was sent.
+	log.initialize({ spyRendererConsole: true });
+
+	log.transports.file.resolvePathFn = () => path.join(app.getPath('logs'), 'main.log');
+	log.transports.file.maxSize = MAX_LOG_SIZE;
+
+	// Redirects the existing console.* calls throughout the main process into
+	// the log without editing their call sites. They keep printing to stdout in
+	// development via the console transport.
+	Object.assign(console, log.functions);
+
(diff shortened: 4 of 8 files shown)


juanmaguitar:
## Windows verification

The PR description above only claimed macOS testing, so recording the Windows result here.

Tested with the packaged installer from the CI build for this branch — `WordPress Contributor Toolkit Setup 0.1.0.exe`, Windows Build #131, commit d371d76.

**Menu — confirmed working.** `Help` renders with `Open App Log`, `Show Logs Folder`, a separator, and `Toggle Developer Tools` (showing its `Ctrl+Shift+I` accelerator). The rebuilt default submenus are all present as `File Edit View Window Help`, with no macOS-only `appMenu` leaking in — which is the part I most wanted eyes on, since setting an application menu replaces Electron's default wholesale and a bad role would have shown up as a missing or empty submenu.

**Still unconfirmed on Windows:** the contents of `main.log` itself, i.e. that child-process output is captured and that the file lands at `%APPDATA%\WordPress Contributor Toolkit\logs\main.log`. That half was verified on macOS only. Given the path comes from `app.getPath('logs')` and the capture code is platform-independent, I have no specific reason to expect a difference — but it has not actually been observed on Windows yet, so treat it as untested rather than passing.

One thing worth a reviewer's attention while testing on Windows: the log will contain the local user profile path (`C:\Users\<name>\...`) in every spawn line. That is by design — it is the same information a bug report needs — but it is the reason the description flags that the log is unredacted.

juanmaguitar:
## Log filename changed (4725d22)

The path in my Windows comment above is now stale. `main.log` was electron-log's default, and it is the wrong name for a file whose entire purpose is to be attached to a bug report — by the time someone reads it, it has been separated from the app-named folder that identified it.

It is now `wordpress-contributor-toolkit-<platform>.log`:

| Platform | Path |
| --- | --- |
| Windows | `%APPDATA%\WordPress Contributor Toolkit\logs\wordpress-contributor-toolkit-windows.log` |
| macOS | `~/Library/Logs/WordPress Contributor Toolkit/wordpress-contributor-toolkit-macos.log` |
| Linux | `~/.config/WordPress Contributor Toolkit/logs/wordpress-contributor-toolkit-linux.log` |

The platform is in the name because one issue thread often collects logs from several contributors, and which OS a log came from is the first thing a reader needs. Friendly names rather than `process.platform`'s own values, since `win32` reads as 32-bit to most people and this string is seen by contributors, not just by code.

Verified locally that both the log file and its rotation sibling get the new name:

```
wordpress-contributor-toolkit-macos.log
wordpress-contributor-toolkit-macos.old.log
```

The menu entries resolve the path through the transport rather than recomputing it, so `Open App Log` and `Show Logs Folder` needed no change and stay correct if the naming ever changes again.

juanmaguitar:
Thanks @Copilot — all three were worth acting on. Fixed in 7c168c3.

**1. `onDone` never called on a spawn failure** — the substantive one, and it was a regression this PR introduced.

Before this PR there was no `'error'` listener on the npm child at all, which means a spawn failure was an unhandled `'error'` event and took the main process down. Adding the listener stopped the crash but created the failure mode described: nothing would settle the request.

I checked the premise before fixing it. On Electron's own Node 20.18.1, a spawn failure emits both events — `error:ENOENT` then `close:-2` — so `onDone` does currently run and the request is not stuck in practice. But Node documents that the exit events "may or may not" fire after an error, this cannot be verified on Windows from here, and a spawn failure is the exact scenario the log exists to expose. It must not also be a silent hang.

Both events now route through a single `settle()`, so `onDone` runs exactly once. The error path is deferred by a turn so `close` — the path that knows about the engines retry — wins whenever it arrives; I measured the ordering rather than assuming it (`error` → `setImmediate` → `close` → `setTimeout(0)`). The retry is additionally guarded on `!settled` so a run already reported as finished cannot start a second attempt behind the caller's back.

Verified against the contract directly, covering both event orderings:

| Case | `onDone` | Retried |
| --- | --- | --- |
| close only (normal exit) | `[0]` | no |
| error then close (real spawn failure) | `[-2]` | no |
| error, close never arrives | `[null]` | no |
| close triggers engines retry | `[]` — attempt 2 settles | yes |
| timer beats close (hypothetical ordering) | `[null]` | no |
| two error events | `[null]` | no |

This lives in `main.js` and so is not reachable from the `node:test` suite, which only covers Electron-free modules — hence the standalone verification rather than a committed test.

**2. `logScope` basename collision** — correct, and a real corruption risk rather than just a mislabeling one. The log's line buffers are keyed by scope, servers for several sites can run concurrently, so two sites sharing a folder name under different parents would have interleaved each other's *partial* lines into corrupted ones. Scope is now suffixed with a short hash of the full path; the full path is still logged once at spawn, so the hash never has to be decoded by a reader.

**3. Stale "four functions" comment** — right, it was six. Reworded to not restate a count that has to be maintained.

PR #51, opened 2026-07-29
Lock the "Install npm dependencies" step until the clone finishes

### LINKED ISSUE #47: "Install npm dependencies" is clickable while the repo is still cloning
### Description

The "Install npm dependencies" button in the site setup checklist is enabled as soon as the site view appears, which happens **before** the `wordpress-develop` clone has finished. Clicking it at that point runs `npm install` against an incomplete working tree and the setup breaks.

Waiting for the clone to finish and then clicking it works fine.

Expected: step 2 stays locked/disabled until the clone completes, and step 1 ("Download WordPress development version") reflects that it is still in progress instead of showing as completed from the start.

1. Open the app and create a new WordPress Core site.
2. As soon as the site detail view appears (immediately — the clone is still running), click "Install npm dependencies".
3. The install runs against a half-cloned tree and the process breaks.

### Operating system

Windows (reproduced on the Buildkite Windows build). The underlying UI bug is cross-platform — it's just easier to hit on Windows because the clone takes longer.

### App version

Windows build from Buildkite, `trunk` @ 668611d.

### Technical notes

- `handleCreateSiteSubmit` (`src/renderer/index.jsx:258-279`) adds the site to state and calls `setActiveSite(targetDir)` **before** `await window.api.setupWordPress(...)`, so the checklist renders while `git.clone` (`src/main.js:427`) is still running.
- The `download` step in `baseSteps` is hardcoded to `done: true, ready: true` (`src/renderer/index.jsx:1369-1376`), and the install button is only disabled by `statusLoading || installing || hasNodeModules` (`src/renderer/index.jsx:1388`) — nothing about the clone.
- The signal already exists but is dropped: `App` passes `isPending={Boolean(pendingSite && pendingSite.targetDir === s)}` to `SiteRow` (`src/renderer/index.jsx:608`), and `SiteRow` never declares that prop (`src/renderer/index.jsx:686`).
- Suggested fix: accept `isPending` in `SiteRow`, derive the `download` step's `done` from it, and add it to the install button's `disabled` plus the install step's `ready`. Make sure `pendingSite` is cleared on every code path (including the error path) so the button can't get stuck disabled.
- `src/renderer/index.js` and `bundle.js` are committed esbuild artifacts — regenerate with `npm run build:once`.

juanmaguitar:
> **Stack — 2 of 5** · `trunk` → #50 → **#51** → #52 → #56 → #58

Fixes #47

## Problem

The site detail view renders as soon as a site is created — **before** `git.clone` of `wordpress-develop` has finished. The checklist showed step 1 ("Download WordPress development version") as already complete and left step 2 ("Install npm dependencies") enabled, so `npm install` could run against a half-cloned working tree and break the setup.

Easiest to hit on Windows because the clone takes longer, but the bug is in the renderer and is cross-platform.

## Fix

While the clone runs, step 1 shows as *current* ("Cloning the WordPress develop repository…") and step 2 renders *locked* with a disabled button.

The signal already existed and was silently dropped: `App` passed `isPending` to `SiteRow`, and `SiteRow` never declared the prop.

## How to test

Run the app from this branch: `npm install && npm start`

1. Click **Create WordPress Core site**, give it a name and a location, submit.
2. **Immediately** — while the clone is still running — look at the checklist:
   - step 1 is *current*, described as "Cloning the WordPress develop repository…"
   - step 2 is *locked* and its button is disabled, so it cannot be clicked
3. Wait for the clone to finish. Step 1 flips to *completed* and step 2's button becomes enabled.
4. Regression check: open a site that was already set up. Its checklist behaves exactly as before — install/build/dev buttons gated only by whether `node_modules` and `build/` exist.

To see the bug this fixes, run the same steps on `trunk`: step 2 is clickable from the first frame, and clicking it during the clone breaks the setup.

Automated: `npm test` (9 new cases covering the full step-state truth table).

<details>
<summary><b>Changes</b></summary>

- **`src/renderer/setup-steps.cjs` (new)** — pure, dependency-free `computeSetupStepState({ isPending, statusLoading, hasNodeModules, hasBuilt, installing, building, starting })` returning `done` / `ready` / `disabled` for the four checklist steps. Extracted so the gating logic can be unit tested without a DOM (the repo has no React test setup); the renderer bundle imports it, `node --test` requires it directly.
- **`src/renderer/index.jsx`** — `SiteRow` accepts `isPending`; all four steps take their state from the helper.
- **Pending state made robust** — `pendingSite` becomes `pendingSites` (array), because on a directory-name collision the main process settles on a different dir than the renderer's optimistic one and the single-value compare missed the row. It is now also cleared in `handleCreateSiteSubmit`'s `finally` — `setupWordPress` resolving *is* the clone finishing — so the button cannot get stuck disabled if the `done` status event is missed.
- **`test/setup-steps.test.cjs` (new)** — 9 cases, including the regression itself and the unchanged truth table for the build/dev buttons.
- `src/renderer/index.js` regenerated with `npm run build:once` (committed esbuild artifact).

</details>

### DIFF
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -30964,6 +30964,44 @@ WARNING: This link could potentially be dangerous`)) {
     }
   });
 
+  // src/renderer/setup-steps.cjs
+  var require_setup_steps = __commonJS({
+    "src/renderer/setup-steps.cjs"(exports, module) {
+      "use strict";
+      function computeSetupStepState2(flags = {}) {
+        const isPending = Boolean(flags.isPending);
+        const statusLoading = Boolean(flags.statusLoading);
+        const hasNodeModules = Boolean(flags.hasNodeModules);
+        const hasBuilt = Boolean(flags.hasBuilt);
+        const installing = Boolean(flags.installing);
+        const building = Boolean(flags.building);
+        const starting = Boolean(flags.starting);
+        return {
+          download: {
+            done: !isPending,
+            ready: true
+          },
+          install: {
+            done: hasNodeModules,
+            ready: !isPending,
+            disabled: isPending || statusLoading || installing || hasNodeModules
+          },
+          build: {
+            done: hasBuilt,
+            ready: hasNodeModules,
+            disabled: statusLoading || building || !hasNodeModules || hasBuilt
+          },
+          dev: {
+            done: false,
+            ready: hasBuilt,
+            disabled: statusLoading || starting || !hasBuilt
+          }
+        };
+      }
+      module.exports = { computeSetupStepState: computeSetupStepState2 };
+    }
+  });
+
   // src/renderer/index.jsx
   var import_react69 = __toESM(require_react());
   var import_client2 = __toESM(require_client());
@@ -53327,6 +53365,7 @@ If there's a particular need for this, please submit a feature request at https:
 
   // src/renderer/index.jsx
   var import_xterm = __toESM(require_xterm());
+  var import_setup_steps = __toESM(require_setup_steps());
   var import_jsx_runtime61 = __toESM(require_jsx_runtime());
   var TERMINAL_ALLOWED_SCRIPTS = ["build", "build:dev", "dev", "test", "watch", "grunt"];
   var TERMINAL_INSTALL_ALIASES = ["npm install", "npm i", "install"];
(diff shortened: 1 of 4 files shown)


juanmaguitar:
Manually verified on Windows, where the original report came from: the "Install npm dependencies" button stays disabled while the `wordpress-develop` clone is running, and only becomes clickable once the clone finishes.

That closes the "not yet done" item in the description — no need for a reviewer to repeat the end-to-end run.

PR #52, opened 2026-07-29
Fix "Run first full build": no output in the terminal and flashing Windows consoles

### LINKED ISSUE #48: "Run first full build" shows no output in the app terminal and flashes Windows console windows
### Description

On Windows, clicking "Run first full build" produces **no output at all** in the app's terminal panel. Instead, black Windows console windows appear and disappear on screen.

Expected: the build output streams into the app's terminal, the same way `npm install` does, and no external console windows are shown.

2. Click "Install npm dependencies" and wait for it to complete (this step works).
4. Nothing appears in the app terminal; one or more console windows flash on screen and vanish.

### Operating system

Windows (Buildkite build). The missing-output half also affects macOS; the console windows are Windows-only.

### App version

Windows build from Buildkite, `trunk` @ 668611d.

### Technical notes

Two independent causes:

**1. No output (cross-platform).** The checklist button calls `onClick={() => runScript('build')}` (`src/renderer/index.jsx:1402`) without `onLog`/`onDone`. `runScript` only forwards output to `appendNpm`, and `npmLogs` (`src/renderer/index.jsx:693`) is never rendered anywhere — `npmRef` is never attached to a DOM node — so the output goes nowhere. Compare with the paths that do work: `runInstallWithTerminal` (`src/renderer/index.jsx:925-933`), the `npm run <script>` terminal command (`src/renderer/index.jsx:1032-1043`) and the dev-server `watch` (`src/renderer/index.jsx:1245-1257`), which all pass `onLog: writeToTerminal`.

**2. Console windows (Windows-only).** `electron.exe` is a GUI-subsystem binary, so `windowsHide: true` on the `spawn` in the `npm:run-script` handler (`src/main.js:569`) doesn't allocate a console for the child. When npm then runs `cmd.exe /d /s /c "grunt build"` (and afterwards the `grunt.cmd` shim), Windows allocates a **new visible** console for each one because there is none to inherit. Since `src/script-runner.js` loads npm's CLI in-process, patching `child_process.spawn`/`exec*` there to force `windowsHide: true` before requiring the npm CLI gives `cmd.exe` a hidden console that its own descendants inherit, which removes all the flashes. The same treatment applies to `src/install-runner.js` and `src/server-runner.js`.

**Related gap worth fixing alongside:** neither `npm:install` nor `npm:run-script` registers `child.on('error')` (`src/main.js:506-590`). If the `spawn` itself fails (ENOENT/EPERM — common on Windows), no `:done` event is ever emitted and the UI stays stuck in "building" with no message at all.

juanmaguitar:
> **Stack — 3 of 5** · `trunk` → #50 → #51 → **#52** → #56 → #58

Fixes #48.

## Problem

Clicking **Run first full build** produced no output at all in the app's terminal. On Windows it also flashed black console windows on and off the screen.

## Fix

Two independent defects:

1. **No output (all platforms).** The checklist button never sent the build's output anywhere that gets rendered. It now streams into the terminal, exactly like `npm install` already did.
2. **Flashing console windows (Windows only).** Every child process in the chain was allocating its own visible console. New `src/hide-child-windows.js` gives them a hidden one that the whole subtree inherits.

## How to test

Needs a site with dependencies already installed. The console-window half is Windows-only — for that, use a packaged build from this branch.

1. Click **Run first full build**.
2. **Output:** grunt's output appears in the Terminal panel *live*, task by task (`Running "clean:files" (clean) task`, `>> 5645 paths cleaned.`, …), not at the end and not nowhere.
3. **Ctrl+C** in the terminal stops the build and returns the prompt.
4. **Windows only:** watch the screen for the whole build. No black console window should appear at any point — not even a flash. On `trunk` they flash repeatedly.
5. Let it finish: the terminal prints the exit code and step 3 flips to *completed*.

To see the bug, run step 1 on `trunk`: the terminal stays empty for the entire build.

Automated: `npm test` (7 new cases in `test/hide-child-windows.test.cjs`).

<details>
<summary><b>1 — Why no output appeared</b></summary>

The checklist button called `runScript('build')` with no `onLog`/`onDone`, so output only reached `appendNpm` → `npmLogs` — a state value that is never rendered, since `npmRef` is never attached to a DOM node. Now it goes through `writeToTerminal`, exactly like `runInstallWithTerminal`, the `npm run <script>` terminal command, and the dev-server `watch` already do.

</details>

<details>
<summary><b>2 — Why Windows opened a console for every child</b></summary>

`electron.exe` is a GUI-subsystem binary, so the runners spawned from `main.js` have no console of their own — `windowsHide: true` there has nothing to inherit. When npm's run-script then spawns `cmd.exe /d /s /c "grunt build"` (and cmd runs the `grunt.cmd` shim), Windows allocates a new **visible** console for each.

The runners load npm's CLI in-process, so `src/hide-child-windows.js` patches `child_process` before that require. `windowsHide: true` maps to `CREATE_NO_WINDOW`, which gives `cmd.exe` a console that is never displayed and that its descendants inherit, so one patch covers the whole subtree. Applied to the script, install, server and playground-web runners; a no-op off Windows.

</details>

<details>
<summary><b>Scope change from stacking: the spawn-error guard now comes from #50</b></summary>

This PR originally carried a third fix — neither `npm:install` nor `npm:run-script` listened for `'error'` on the child, so a spawn that failed outright (ENOENT/EPERM) could leave the button busy forever.

Stacking on #50 revealed that it had grown its own guard for the same failure: this PR had `finished` / `finish()`, #50 has `settled` / `settle()`. #50's is the better one and it won — it defers by a turn so `close` (the path that knows about the engines retry) wins whenever it does arrive, and it guards the retry with `!settled` so a run the caller already considers over cannot be restarted behind its back. The duplicate here was removed.

What was kept from this PR is the part #50 didn't have: the `onLog('stderr', 'Failed to start: …')` line. #50 only wrote the failure to the log file; the person who just clicked the button needs to see it in the terminal too.

</details>

### DIFF
--- a/src/hide-child-windows.js
+++ b/src/hide-child-windows.js
@@ -1,0 +1,74 @@
+// Keeps Windows from flashing a console window for every grandchild process.
+//
+// electron.exe is a GUI-subsystem binary, so the runners we spawn from main.js
+// have no console of their own — `windowsHide: true` there only stops Windows
+// from showing one, it doesn't create one to inherit. When npm's run-script
+// then spawns `cmd.exe /d /s /c "grunt build"` (and cmd in turn runs the
+// grunt.cmd shim), each of those console applications finds nothing to inherit
+// and Windows allocates a brand new *visible* console for it.
+//
+// The runners load npm's CLI in-process, so patching child_process here — before
+// requiring that CLI — reaches those spawns. `windowsHide: true` maps to
+// CREATE_NO_WINDOW, which gives cmd.exe a console that is never displayed and
+// which its own descendants inherit, so one patch covers the whole subtree.
+
+const PATCHED = Symbol.for('wp-dev-env.windowsHidePatched');
+
+const METHODS = [
+	'spawn',
+	'spawnSync',
+	'exec',
+	'execSync',
+	'execFile',
+	'execFileSync',
+	'fork'
+];
+
+function isPlainObject(value) {
+	return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+// These functions have incompatible signatures — spawn(cmd, args, opts),
+// exec(cmd, opts, cb), execFile(file, args, opts, cb) — so instead of special
+// casing each one we find the options object positionally: it is the last
+// argument that is a plain object, ignoring any trailing callback. spawn's
+// `args` is an array, so it is never mistaken for options.
+function withWindowsHide(args) {
+	const next = args.slice();
+	let index = next.length - 1;
+	while (index >= 0 && typeof next[index] === 'function') index -= 1;
+
+	if (index >= 0 && isPlainObject(next[index])) {
+		next[index] = { ...next[index], windowsHide: true };
+		return next;
+	}
+
+	next.splice(index + 1, 0, { windowsHide: true });
+	return next;
+}
+
+// Forces `windowsHide: true` on every child_process entry point of `cp`.
+// A no-op off Windows, and idempotent so requiring this twice can't double-wrap.
+function patchChildProcess(cp, platform = process.platform) {
+	if (platform !== 'win32') return cp;
+	if (cp[PATCHED]) return cp;
+
+	for (const name of METHODS) {
+		const original = cp[name];
+		if (typeof original !== 'function') continue;
+		cp[name] = function patched(...args) {
+			return original.apply(this, withWindowsHide(args));
+		};
+	}
+
+	Object.defineProperty(cp, PATCHED, { value: true, enumerable: false });
+	return cp;
+}
+
+// Patches the live child_process module for this process and everything it
+// requires afterwards.
+function hideChildWindows() {
+	return patchChildProcess(require('child_process'));
+}
+
+module.exports = { patchChildProcess, hideChildWindows, withWindowsHide };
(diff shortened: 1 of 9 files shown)


juanmaguitar:
## Note from stacking: overlap with #50 resolved here

This PR is now based on #51 (and through it, #50), and merging that chain in surfaced a real overlap that was invisible while the branches were parallel.

Both this PR and #50 had grown their own guard against the same failure — a spawn that emits `error` and may never emit `close`, leaving the renderer without a `:done` event and the UI stuck. This PR had `finished` / `finish()`; #50 had `settled` / `settle()`.

#50's version is the better one and it won: it defers by a turn so that `close` — the path that knows about the engines retry — wins whenever it does arrive, and it guards the retry with `!settled` so a run the caller already considers over cannot be restarted behind its back. This PR's duplicate was removed.

What was kept from this PR is the part #50 didn't have: the `onLog('stderr', 'Failed to start: …')` line. #50 only wrote the failure to the log file; the person who just clicked the button needs to see it in the app's terminal too.

Net effect on this PR's scope: the `child.on('error')` machinery is no longer its contribution — the terminal output and the Windows console-window fix are.

Issue #54, opened 2026-07-29
`npm run build` re-runs `npm install` and fails with EBADENGINE because relaxed engine checks are not applied to scripts


juanmaguitar:
### Description

The first thing `npm run build` does is a nested `npm install`, and that install fails with `EBADENGINE` — the exact failure the app already works around for its own "Install npm dependencies" step (#37 / #38).

It is not fatal (the caller swallows the error and grunt continues), but it costs ~20 seconds on every build and dumps a full `Unsupported engine` stack trace plus a raw error-object dump into the log. Once #48 lands and build output is actually visible in the app terminal, this becomes the first thing a Contributor Day newcomer sees after clicking "Run first full build".

### Step-by-step reproduction instructions

1. Create a site and wait for the clone to finish.
2. Click "Install npm dependencies". It fails once with `EBADENGINE`, then succeeds on the automatic relaxed-engines retry (working as designed, per #38).
3. Click "Run first full build".
4. The build prints `package.json has been modified. Installing...`, then a full `npm error code EBADENGINE` block for `@eslint/compat@2.1.0`, then `Error: Command failed: npm install` with a stack trace, and only then reaches `Running "build" task`.

### Operating system

Reproduced on Windows 11 (10.0.26100). The cause is platform-independent — it should reproduce anywhere Electron's bundled Node is older than what a dependency requires.

### App version

`electron-setup-wordpress-core` 0.1.0 · electron 32.3.3 · node 20.18.1 · production build.

### Technical notes

`wordpress-develop`'s `Gruntfile.js:130` runs, at load time:

```js
// First do `npm install` if package.json has changed.
installChanged.watchPackage();
```

`install-changed` shells out with `execSync('npm install')` (`node_modules/install-changed/lib/main.js:50`), so the build spawns a *second* npm install as a grandchild of the script runner. Observed stack:

```
Error: Command failed: npm install
npm error code EBADENGINE
npm error notsup Required: {"node":"^20.19.0 || ^22.13.0 || >=24"}
npm error notsup Actual:   {"npm":"10.9.3","node":"v20.18.1"}
    at execSync (node:child_process:978:15)
    at Object.watchFile [as watchPackage] (…\node_modules\install-changed\lib\main.js:50:7)
    at module.exports (…\Gruntfile.js:130:17)
```

[...]

Issue #53, opened 2026-07-29
Windows build fails: `grunt gutenberg:verify` aborts with `spawn EINVAL`


juanmaguitar:
### Description

On Windows, `npm run build` never completes. Grunt aborts on its very first task with:

```
Running "build" task

Running "gutenberg:verify" task
Warning: spawn EINVAL Use --force to continue.

Aborted due to warnings.
npm verbose exit 3
```

The build exits with code 3 and no `build/` output is produced, so the wizard can never reach a working site. Until #48 is fixed the failure is completely invisible in the app — this was only found by reading the on-disk log added in #50.

### Step-by-step reproduction instructions

1. Create a site and wait for the clone to finish.
2. Click "Install npm dependencies" and wait for it to complete.
3. Click "Run first full build".
4. The build exits with code 3 at `gutenberg:verify`.

### Operating system

Windows 11 (10.0.26100), x64. Not reproducible on macOS.

### App version

`electron-setup-wordpress-core` 0.1.0 · electron 32.3.3 · node 20.18.1 · production build.

### Technical notes

Root cause chain:

1. `wordpress-develop`'s `Gruntfile.js` (`gutenberg:verify`, ~L1679) spawns Node directly:

   ```js
   grunt.util.spawn( {
       cmd: 'node',
       args: [ 'tools/gutenberg/utils.js' ],
       opts: { stdio: 'inherit' }
   }, ... );
   ```

   `grunt.util.spawn` (grunt-legacy-util) goes to `child_process.spawn` with `shell: false`.

2. There is no real `node` binary on the host — that is the whole point of this app. The only `node` on `PATH` is the shim written by `ensureNodeShimDir()` in `src/main.js:34-35`, and on Windows it is a **`.cmd` / `.bat`** file:

   ```js
   const content = `@echo off\r\nset ELECTRON_RUN_AS_NODE=1\r\n"${process.execPath}" %*\r\n`;
   fs.writeFileSync(path.join(nodeShimDir, 'node.cmd'), content);
   fs.writeFileSync(path.join(nodeShimDir, 'node.bat'), content);
   ```

   `src/main.js:49-50` explains why no `node.exe` is emitted: Electron's exe depends on adjacent DLLs, so copying it to a temp dir yields `STATUS_DLL_NOT_FOUND`.

[...]

PR #55, opened 2026-07-29
[DO NOT MERGE] Windows fixes integration branch (#50 + #51 + #52 + #56 + #58)

juanmaguitar:
## ⚠️ Do not merge — testing branch only

This PR exists so Buildkite produces a **single Windows artifact containing all the in-flight Windows fixes at once**, so they can be tested together on a Windows VM. Each fix keeps its own PR for review and merges separately; this branch is disposable and will be rebuilt as more PRs land.

## What's in it

`trunk` with these branches merged in:

| PR | Issue | Fix |
| --- | --- | --- |
| #50 | #49 | Write an app log to disk and add a Help menu to open it |
| #51 | #47 | Lock the "Install npm dependencies" step until the clone finishes |
| #52 | #48 | "Run first full build": stream output to the terminal, stop the Windows console flashes |
| #56 | #53 | Windows build aborting at `grunt gutenberg:verify` with `spawn EINVAL` |
| #58 | #54 | Relax npm engine checks for scripts so nested installs don't fail |

`npm test` passes on the merge result: **70/70** (including #58's two integration tests, which run real npm).

## Conflict resolution

Every conflict was additive — two branches adding different things to the same function or JSX element. Both sides were kept; no behaviour was dropped.

**`src/main.js` — `runNpmWithEngineRetry`**, touched by #50, #52, #56 and #58:

- Signature: `logScope` (#50) **and** `relaxEnginesFromStart` (#58) both kept. #58 branched off `trunk` without #50, so its version had dropped `logScope`.
- The two `child.on('error')` handlers (#50 and #52) folded into one that writes to the log file *and* reports the failure to the renderer via `finish(-1)`.
- In `child.on('close')`, the exit is logged (#50) before the `if (finished) return` guard (#52), so an exit is recorded even when the run already ended through the error path.
- Comment block: kept #58's rewrite, which documents both knobs (`retryOnEngineMismatch` vs `relaxEnginesFromStart`) and supersedes the older one.
- In the `npm:run-script` handler, both `logScope` (#50) and `relaxEnginesFromStart: true` (#58) are passed.

**`src/renderer/index.jsx`** — the build button uses `onClick={runBuildWithTerminal}` (#52) **and** `disabled={stepState.build.disabled}` (#51).

**`src/renderer/index.js`** — regenerated with `npm run build:once` rather than hand-merged.

#56 merged cleanly on top of the rest.

## What only exists on this branch

#52 and #56 both patch `child_process`, by different routes, and this is the first place they run together. #56's patch is preloaded into every descendant Node process via `NODE_OPTIONS=--require`, so it lands first; #52's `hide-child-windows` is applied inside the runners and wraps the already-rewritten `spawn`. The orders compose — target rewrite first, `windowsHide` on top — but if something behaves oddly on Windows, that interaction is the first place to look.

## How to refresh it as more PRs land

```
git checkout juanmaguitar/windows-fixes-integration
git merge origin/trunk
git merge origin/<new-pr-branch>
npm run build:once && npm test
git push
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@wp-playground/cli": "^3.0.8",
         "@xterm/xterm": "^5.5.0",
         "diff": "^5.2.0",
+        "electron-log": "^5.4.4",
         "electron-store": "^10.0.1",
         "extract-zip": "^2.0.1",
         "fs-extra": "^11.2.0",
@@ -5138,6 +5139,15 @@
       },
       "engines": {
         "node": ">=12"
+      }
+    },
+    "node_modules/electron-log": {
+      "version": "5.4.4",
+      "resolved": "https://registry.npmjs.org/electron-log/-/electron-log-5.4.4.tgz",
+      "integrity": "sha512-istWgaXjBfURBSS8LWVW9C3jsc6+ac+tY1lXrQEOTp0lVj+a4OlO1Tmqb36GgnEUDv92DGC9VI1HNXwJinWpgA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">= 14"
       }
     },
     "node_modules/electron-publish": {
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@wp-playground/cli": "^3.0.8",
     "@xterm/xterm": "^5.5.0",
     "diff": "^5.2.0",
+    "electron-log": "^5.4.4",
     "electron-store": "^10.0.1",
     "extract-zip": "^2.0.1",
     "fs-extra": "^11.2.0",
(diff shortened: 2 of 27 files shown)


juanmaguitar:
## Superseded by the stack

The five PRs this branch was bundling are now a stack, each based on the previous one:

`trunk` → #50 → #51 → #52 → #56 → #58

That makes this branch redundant in both of its jobs:

- **Combined artifact** — #58 is the tip of the stack, so its branch already contains all five fixes. Buildkite building #58 produces the same Windows `.exe` this branch was built for.
- **Conflict resolution** — the conflicts between #50/#51/#52/#58 were resolved only here, on a branch nobody would merge. They now live inside the PRs themselves, reviewed as part of the change that causes them.

Closing. The Windows validation done from this branch's artifact still stands and is referenced from each PR in the stack.

PR #56, opened 2026-07-29
Fix Windows build aborting at `grunt gutenberg:verify` with `spawn EINVAL`

### LINKED ISSUE #53: Windows build fails: `grunt gutenberg:verify` aborts with `spawn EINVAL`

juanmaguitar:
> **Stack — 4 of 5** · `trunk` → #50 → #51 → #52 → **#56** → #58

Fixes #53.

## Problem

On Windows `npm run build` died on its very first grunt subtask (`gutenberg:verify`) with `spawn EINVAL`, exit code 3, and no `build/` output — so the wizard could never reach a working site.

`wordpress-develop`'s Gruntfile spawns a bare `node`. There is no real `node` on the host — that's the whole point of this app — so it hits our `node.cmd` shim, and Node ≥ 20.12.2 refuses to spawn a `.cmd` without `shell: true`.

## Fix

Grunt runs two levels below us (`script-runner` → `cmd.exe` → `grunt.cmd` → node), so patching `child_process` in our own process can't reach it. The patch is preloaded into **every** descendant Node process via `NODE_OPTIONS=--require`, and rewrites `node` to Electron's own executable — no shell involved.

## How to test

**Windows only** — the failure does not reproduce on macOS or Linux. Use a packaged build from this branch (Buildkite on this branch produces one).

1. Create a site, let the clone finish, click **Install npm dependencies** and wait for it to complete.
2. Click **Run first full build**.
3. It should get past `gutenberg:verify` and `gutenberg:download` and run to completion, producing a `build/` directory. On `trunk` it dies on the first of those with `spawn EINVAL` and exit code 3.
4. Open **Help → Open App Log** and search it for:
   - `EINVAL` — should not appear
   - `MODULE_NOT_FOUND` / `win-spawn-patch.js` — should not appear either; that was the bug fixed in 9fb73e4, where the preload path arrived with its backslashes stripped and killed every child process
5. `exited with code 0` for the `build#…` scope.

Automated: `npm test` (11 cases in `test/win-spawn-patch.test.cjs`, plus three `buildChildEnv` cases).

<details>
<summary><b>The full failure chain</b></summary>

1. `wordpress-develop`'s Gruntfile does `grunt.util.spawn({ cmd: 'node', args: [...], opts: { stdio: 'inherit' } })`, which reaches `child_process.spawn` with `shell: false`.
2. The only `node` on `PATH` is the shim from `ensureNodeShimDir()`, and on Windows it's a `.cmd`/`.bat` — there is deliberately no `node.exe`, because Electron's exe needs its adjacent DLLs.
3. Node >= 20.12.2 (CVE-2024-27980) refuses to spawn a `.bat`/`.cmd` without `shell: true`. Electron 32 bundles Node 20.18.1 → `EINVAL`.

`gutenberg:download` does the same thing and would have failed next.

</details>

<details>
<summary><b>What the patch rewrites, and the changes</b></summary>

- `node` → Electron's own `process.execPath` with `ELECTRON_RUN_AS_NODE=1`. **No shell**, so there is no quoting hazard at all — this is the branch that fixes the reported failure.
- `npm` / `npx` → `execPath` + their JS CLIs, same way.
- any other `.cmd`/`.bat` shim (package bin stubs) → `shell: true` with cmd-safe quoting, as a last resort.

Electron only honours `NODE_OPTIONS` in packaged builds when `ELECTRON_RUN_AS_NODE` is set, which `buildChildEnv` already does and all our children inherit.

Changes:

- **`src/win-spawn-patch.js`** (new) — the preloaded patch. Self-contained (Node built-ins only) because it's copied into the temp shim dir: `--require` into a path inside `app.asar` is not reliable under `ELECTRON_RUN_AS_NODE`. Its core is a pure, platform-parameterised `resolveSpawnTarget()`, and it only self-applies when `WPTK_SPAWN_PATCH=1`, so requiring it from a test never mutates the test process.
- **`src/main.js`** — `ensureNodeShimDir()` copies the patch next to the existing `node.cmd`/`npm.cmd` shims and records the npm/npx CLI paths; both are handed to `buildChildEnv`. Copy failure stays non-fatal — worst case we're exactly where we are today.
- **`src/npm-runner.js`** — `buildChildEnv` **appends** `--require "<patch>"` to any inherited `NODE_OPTIONS` (never replaces it) and sets `WPTK_SPAWN_PATCH` / `WPTK_NPM_CLI` / `WPTK_NPX_CLI`. Inert on POSIX and when no patch path is supplied.

</details>

<details>
<summary><b>The bug this shipped with, and how it was caught (commit 9fb73e4)</b></summary>

The first Windows test of this branch failed immediately, and took every child process with it:

```
Error: Cannot find module 'C:UsersJuanMaAppDataLocalTempelectron-node-shims-9832win-spawn-patch.js'
Require stack: - internal/preload
```

Every backslash is gone. Node doesn't read `NODE_OPTIONS` literally — it tokenises it with shell-like rules, and inside quotes a backslash is an escape character. The preload path arrived mangled, `MODULE_NOT_FOUND`, and the process died before running a line. Not just the build: everything goes through `buildChildEnv`.

Fixed by passing the path with forward slashes, which Windows resolves fine in a `require` and which — unlike doubling the backslashes — doesn't depend on how many layers of unescaping the value passes through.

Worth knowing for review: the original test asserted the exact string `--require "C:\shims\win-spawn-patch.js"`, so it passed *because* it encoded the broken value. A string test can't know what Node's tokeniser does with that string afterwards. The assertion now pins the property that matters — no backslash survives in the flag.

</details>

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -24,6 +24,12 @@ const WORDPRESS_GIT_URL = 'https://github.com/WordPress/wordpress-develop.git';
 
 // Provide a PATH shim so npm's spawned scripts can find a 'node' binary that maps to Electron's Node
 let nodeShimDir = null;
+// Windows-only: absolute path of the child_process patch copied next to the
+// shims, preloaded into descendant Node processes via NODE_OPTIONS so that a
+// bare spawn('node') hitting node.cmd does not fail with EINVAL.
+let spawnPatchPath = null;
+let npmCliPath = null;
+let npxCliPath = null;
 function ensureNodeShimDir() {
     if (nodeShimDir) return nodeShimDir;
     nodeShimDir = path.join(os.tmpdir(), `electron-node-shims-${process.pid}`);
@@ -39,12 +45,23 @@ function ensureNodeShimDir() {
                 const npmRootDir = path.dirname(npmPkgJsonPath);
                 const npmCliAbsPath = path.join(npmRootDir, 'bin', 'npm-cli.js');
                 const npxCliAbsPath = path.join(npmRootDir, 'bin', 'npx-cli.js');
+                npmCliPath = npmCliAbsPath;
+                npxCliPath = npxCliAbsPath;
                 const npmCmd = `@echo off\r\nset ELECTRON_RUN_AS_NODE=1\r\n"${process.execPath}" "${npmCliAbsPath}" %*\r\n`;
                 const npxCmd = `@echo off\r\nset ELECTRON_RUN_AS_NODE=1\r\n"${process.execPath}" "${npxCliAbsPath}" %*\r\n`;
                 fs.writeFileSync(path.join(nodeShimDir, 'npm.cmd'), npmCmd);
                 fs.writeFileSync(path.join(nodeShimDir, 'npm.bat'), npmCmd);
                 fs.writeFileSync(path.join(nodeShimDir, 'npx.cmd'), npxCmd);
                 fs.writeFileSync(path.join(nodeShimDir, 'npx.bat'), npxCmd);
+            } catch {}
+            // Copy the child_process patch out of the app bundle: it is loaded with
+            // --require by child Node processes, and a path inside app.asar is not
+            // reliably resolvable under ELECTRON_RUN_AS_NODE. A failure here only
+            // means we are back to the pre-patch behaviour, so it stays non-fatal.
+            try {
+                const dest = path.join(nodeShimDir, 'win-spawn-patch.js');
+                fs.copyFileSync(path.join(__dirname, 'win-spawn-patch.js'), dest);
+                spawnPatchPath = dest;
             } catch {}
             // Intentionally do NOT create node.exe here, as Electron's exe depends on adjacent DLLs.
             // Using node.exe from a temp dir causes STATUS_DLL_NOT_FOUND (0xC0000135) when spawned by npm.
@@ -529,6 +546,9 @@ function runNpmWithEngineRetry({ runnerPath, args, cwd, onLog, onDone, register,
 			cwd,
 			env: buildChildEnv({
 				shimDir: ensureNodeShimDir(),
+				spawnPatchPath,
+				npmCliPath,
+				npxCliPath,
 				extraEnv: relaxEngines ? RELAXED_ENGINES_ENV : {}
 			}),
 			shell: false,
--- a/src/npm-runner.js
+++ b/src/npm-runner.js
@@ -60,12 +60,21 @@ function shouldRetryWithRelaxedEngines({ code, signal, sawEngineMismatch, alread
 
 // The env block for a spawned npm runner. On Windows both PATH and Path are set
 // and PATHEXT is extended so child npm processes can find the node shim.
+//
+// On Windows the node shim is a .cmd, which Node >= 20.12.2 refuses to spawn
+// without `shell: true` — so `spawnPatchPath` is preloaded into every descendant
+// Node process via NODE_OPTIONS to rewrite those spawns. See win-spawn-patch.js.
+// Electron only honours NODE_OPTIONS when ELECTRON_RUN_AS_NODE is set, which it
+// is just below.
 function buildChildEnv({
 	shimDir,
 	extraEnv = {},
 	baseEnv = process.env,
 	platform = process.platform,
-	execPath = process.execPath
+	execPath = process.execPath,
+	spawnPatchPath = null,
+	npmCliPath = null,
+	npxCliPath = null
 } = {}) {
 	const isWindows = platform === 'win32';
 	const separator = isWindows ? ';' : ':';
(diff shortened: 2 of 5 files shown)


juanmaguitar:
## Caught on Windows: the preload path was arriving mangled

Tested from the integration build (#55) on a Windows VM. Clicking **Install npm dependencies** failed immediately. From the app log (which #50 gave us — without it this would have looked like "nothing happens" again):

```
[2026-07-29 12:31:51.886] [info]  (install#ed21) spawn install-runner.js C:\Users\JuanMa\Desktop\demo1 in C:\Users\JuanMa\Desktop\demo1
[2026-07-29 12:31:52.869] [warn]  (install#ed21) Error: Cannot find module 'C:UsersJuanMaAppDataLocalTempelectron-node-shims-9832win-spawn-patch.js'
[2026-07-29 12:31:52.869] [warn]  (install#ed21) Require stack:
[2026-07-29 12:31:52.871] [warn]  (install#ed21) - internal/preload
[2026-07-29 12:31:52.871] [warn]  (install#ed21)     at Module._resolveFilename (node:internal/modules/cjs/loader:1232:15)
...
[2026-07-29 12:31:52.878] [warn]  (install#ed21)   code: 'MODULE_NOT_FOUND',
[2026-07-29 12:31:52.878] [warn]  (install#ed21)   requireStack: [ 'internal/preload' ]
[2026-07-29 12:31:52.908] [info]  (install#ed21) exited with code 1
```

Every backslash is gone: `C:\Users\JuanMa\AppData\...` reached the preloader as `C:UsersJuanMaAppData...`.

**Cause.** Node doesn't read `NODE_OPTIONS` literally — it tokenises it with shell-like rules, and inside quotes a backslash is an escape character. So `--require "C:\Users\..."` loses its separators, the preload fails, and the process dies before running a single line. Not just the install: everything goes through `buildChildEnv`, so `npm run build` and the Playground server were broken the same way. The branch as it stood took the app down on Windows.

**Fix** (9fb73e4): pass the path with forward slashes. Windows resolves them fine in a `require`, and unlike doubling the backslashes it doesn't depend on how many layers of unescaping the value passes through. The quotes stay, for paths containing spaces.

**Why the tests didn't catch it.** They asserted the exact string `--require "C:\shims\win-spawn-patch.js"` — i.e. the assertion passed *because* it encoded the broken value. A pure string test can't know what Node's tokeniser does with that string afterwards. Added a regression test that pins the property that actually matters (no backslash survives in the flag) instead of the literal.

Ready for a re-test — I'll refresh #55 so the next Windows artifact includes this.

Issue #57, opened 2026-07-29
First build after `Install npm dependencies` pays a redundant nested `npm install` (install-changed's packagehash.txt is never seeded)


juanmaguitar:
### Description

After "Install npm dependencies" succeeds, the **first** "Run first full build" still spends ~20s on a redundant nested `npm install` that has nothing to do. `wordpress-develop`'s Gruntfile runs `install-changed` at load time, and `install-changed` decides whether to install by comparing a hash file, `packagehash.txt`, that our own install path never writes. So the first build always sees "no hash → must install", re-resolving a dependency tree that is already fully installed on disk.

This is a small, one-per-site cost, not a per-build one — see the correction below.

### Correcting the aside in #54

#54's technical notes say:

> `install-changed` fires at all because it compares a package.json hash that our own `npm install` never writes, so it will re-run on every single build even when nothing changed.

The *reason* given there is wrong, and so is "every single build". The recurrence was caused by the `EBADENGINE` failure itself, not by our install not seeding the hash. In `install-changed` the hash write is **inside the `try`, after the install**:

```js
// node_modules/install-changed/lib/main.js:48-55  (install-changed@1.1.0)
try {
  execSync('npm install')                       // :50
  fs.writeFileSync(packageHashPath, recentDigest) // :51 — only reached on success
}
catch (error) {
  console.log(error)                            // swallowed; hash never written
}
```

So while the nested install failed on engines, `packagehash.txt` was never written and every build re-ran it. Once #54's fix lands the nested install *succeeds*, the hash gets written, and subsequent builds print `package.json has not been modified.` and skip the install entirely.

Verified empirically against a synthetic project (local dep pinned to `engines.node: ">=99.0.0"`, `.npmrc` with `engine-strict = true`, build script calling `installChanged.watchPackage()`):

| run | env | output | `packagehash.txt` |
|---|---|---|---|
| build 1 | `engine_strict=false` | `package.json has been modified. Installing...` | written |
| build 2 | `engine_strict=false` | `package.json has not been modified.` | reused, install skipped |
| build | strict (the #54 state) | `package.json has been modified. Installing...` + EBADENGINE | **not** written |

So #54's fix already removes the per-build repeat. What is left is only the first build after a fresh install.

### Step-by-step reproduction instructions

1. Create a site and wait for the clone to finish.
2. Click "Install npm dependencies" and let it complete.
3. Click "Run first full build".
[...]

Issue #46, opened 2026-07-29
Windows 11, v0.1.0 - `npm install` fails with `EBADENGINE` — bundled Node v20.18.0 is below `@eslint/compat`'s required range


juanmaguitar:
In Windows 11, v0.1.0, Site setup completes, but the `npm install` step that follows fails immediately with `EBADENGINE`. The app's bundled Node is **v20.18.0**, while `@eslint/compat@2.1.0` declares `engines.node: "^20.19.0 || ^22.13.0 || >=24"`. It misses the first range by a single patch version.

The failure happens during `Arborist.buildIdealTree`, before anything is written — so **no dependencies are installed at all**, not just the incompatible one.

[See video](https://www.youtube.com/watch?v=jmqM1jhwYMo)

### Environment

| | |
|---|---|
| App | WordPress Contributor Toolkit (`electron-setup-wordpress-core`) v0.1.0 |
| OS | Windows_NT 10.0.26100 |
| Bundled Node | v20.18.0 |
| Bundled npm | 10.9.3 |
| Project dir | `C:\Users\<user>\Desktop\demo-windows` |

### Steps to reproduce

1. Run site setup to completion.
2. Let the app run its `npm install` step.
3. Install exits with code 1.

### Expected

Dependencies install successfully, or the app surfaces an actionable message about the Node version rather than a raw npm stack trace.

### Actual

```
npm verbose stack Error: Unsupported engine
npm verbose stack     at checkEngine (...\npm-install-checks\lib\index.js:16:25)
npm verbose stack     at #checkEngineAndPlatform (...\arborist\lib\arborist\build-ideal-tree.js:200:13)
npm verbose stack     at Arborist.buildIdealTree (...\arborist\lib\arborist\build-ideal-tree.js:184:41)
npm verbose stack     at async Promise.all (index 1)
npm verbose stack     at async Arborist.reify (...\arborist\lib\arborist\reify.js:131:5)
npm verbose pkgid @eslint/compat@2.1.0
npm error code EBADENGINE
npm error engine Unsupported engine
npm error engine Not compatible with your version of node/npm: @eslint/compat@2.1.0
npm error notsup Required: {"node":"^20.19.0 || ^22.13.0 || >=24"}
npm error notsup Actual:   {"npm":"10.9.3","node":"v20.18.0"}
npm verbose exit 1
npm install exited with code 1
```

### Analysis

[...]

PR #40, opened 2026-07-27
Upgrade Electron and bundled Node runtime

ashfame:
Fixes #37.

This is a draft alternative to #38. Instead of retrying installs with Node engine checks relaxed, it updates the application runtime so current `wordpress-develop` dependencies run on a supported Node version.

## Changes

- Upgrade Electron from 32.3.3 to 43.2.0, which bundles Node 24.18.0.
- Upgrade the build-time Node version from 20.19.5 to 24.18.0.
- Upgrade `@wp-playground/cli` to 3.1.47 so its native filesystem dependency can be rebuilt for Electron 43.
- Upgrade `electron-builder` to 26.15.3 and migrate the Windows signing options to its current `signtoolOptions` schema.
- Regenerate the npm lockfile and update the lifecycle-script allowlist.

## Why

Electron 32 bundles Node 20.18.1, while current `wordpress-develop` dependencies require `^20.19.0 || ^22.13.0 || >=24` and enforce that requirement with `engine-strict`. Upgrading the bundled runtime fixes the incompatibility without bypassing package engine declarations.

## Validation completed on Linux

- Clean `npm ci` passed on Ubuntu 24.04 with GCC 13 and Debian Trixie with GCC 14.
- `npm test` passed: 13/13.
- `npm run build:once` passed.
- Linux packaging produced an AppImage and Snap.
- The packaged application opened a real GUI window under Xvfb.
- The packaged runtime reported Electron 43.2.0, Node 24.18.0, and Chromium 150.0.7871.129.
- The packaged Playground CLI and rebuilt native filesystem module loaded successfully.
- The packaged `isomorphic-git` stack cloned current `wordpress-develop/trunk`.
- A real `wordpress-develop` checkout installed 2,567 packages with its normal `engine-strict` setting and no fallback.
- The full WordPress Grunt build completed.
- The Playground server reached ready state; the homepage, wp-admin routing, and Adminer responded successfully.
- The packaged patch generator handled tracked and untracked text changes and skipped binary content.
- The npm lifecycle path was exercised without a host Node installation.

## Draft validation goals

Buildkite should now tell us whether the updated native dependency, Electron packaging, Windows signing, and macOS signing/notarization work on the real platform agents. After green builds, the Windows and macOS artifacts still need a short launch and WordPress workflow smoke test.

The native dependency did not compile with GCC 12, but passed with GCC 13 and GCC 14. Linux builders therefore need Python, `make`, and a modern C++ compiler.

PR #58, opened 2026-07-29
Relax npm engine checks for scripts so nested installs don't fail

### LINKED ISSUE #54: `npm run build` re-runs `npm install` and fails with EBADENGINE because relaxed engine checks are not applied to scripts

juanmaguitar:
> **Stack — 5 of 5** · `trunk` → #50 → #51 → #52 → #56 → **#58**

Fixes #54.

## Problem

The first thing `npm run build` does is a nested `npm install` — `wordpress-develop`'s Gruntfile calls `install-changed` at load time, which shells out with `execSync('npm install')`.

That grandchild inherited strict engine checks, so it hit the same `EBADENGINE` wall the app already knows how to get past, dumping an `Unsupported engine` block and a `Command failed: npm install` stack trace into the log before grunt even started. Not fatal — the caller swallows it — but it's the first thing a Contributor Day newcomer sees now that #48 makes build output visible.

## Fix

Relax engine checks **from the start** for `npm:run-script`, rather than retrying. `npm:install` is untouched: `relaxEnginesFromStart` defaults to `false`, so the #38 retry path and its "show the real reason first" output are unchanged.

## How to test

Best seen on Windows, where Electron's bundled Node (20.18.1) is older than what several dependencies ask for. Needs a site with dependencies already installed.

1. Click **Run first full build**.
2. Open **Help → Open App Log** and follow the `build#…` lines. The log should go from `package.json has been modified. Installing...` straight into the grunt tasks.
3. Neither of these should appear anywhere in that section — on `trunk` both do:
   - `npm error code EBADENGINE` / `Unsupported engine`
   - `Command failed: npm install`
4. Run **Run first full build** a second time. It should now print `package.json has not been modified.` and skip the nested install entirely — that repeat was a side effect of the failure, not a separate bug.
5. Regression check on the install path: click **Install npm dependencies** on a fresh site. It must still try strict first and only then retry — you should see the `EBADENGINE` block followed by `⚠ This site requires a newer Node.js than this app bundles. Retrying with engine checks relaxed…`. That behaviour from #38 is deliberately untouched here.

Automated: `npm test` (two new integration tests that run real npm).

<details>
<summary><b>Why baseline-relax here and retry there</b></summary>

The two are opposite trades, and this one is safe for precisely the reason the retry isn't: nothing is ever restarted.

Scripts must not be retried — they can fail partway through with side effects already on disk, and npm downgrades EBADENGINE to a warning when engine-strict is off, so a warning plus an unrelated non-zero exit would wrongly restart a half-finished script. Baseline-relax sidesteps all of that: engine checks are simply off from the first process onward, and nested installs inherit it.

**Rejected:** relaxing engines inside `buildChildEnv` for all children. It's only called from `runNpmWithEngineRetry`, so that would also relax `npm:install`'s first attempt and destroy that deliberate behaviour.

</details>

<details>
<summary><b>Why the tests have to be integration tests</b></summary>

`test/nested-install.integration.test.cjs` + `test/fixtures/nested-install/`, modelled on the existing `npm-install.integration.test.cjs`. One test pins the bug (strict env → real EBADENGINE in the log, dependency not installed, build still "succeeds" — the noise-not-failure symptom), the other pins the fix.

What's being proven is that `npm_config_engine_strict` survives *two* levels of npm: the outer `npm run` re-exports its own resolved config into the script's environment before the grandchild install reads it. No unit test over `buildChildEnv`'s output strings can show that; only real npm can.

Two fixture details worth knowing:

- The dependency is pinned to `engines.node: ">=99.0.0"` rather than a registry package, so the test can't quietly go green once CI's Node catches up.
- The build script ships as `run-build.js.txt` and is renamed on copy, because `node --test` runs every `.js` under `test/` and was otherwise executing the fixture as a test file (a stray 10s "passing test" that ran a real npm install).

**The tests don't cover the wiring in `main.js`.** They exercise `buildChildEnv` with the flag set explicitly; `main.js` can't be imported under `node --test` because it requires `electron`. So they prove the mechanism, not that the handler passes the flag — that line was confirmed manually on Windows instead.

</details>

<details>
<summary><b>This also fixes the per-build repeat (#54's notes had the cause wrong)</b></summary>

`install-changed` writes its `packagehash.txt` *inside the try, after* the install succeeds (`lib/main.js:50-51`), so while the nested install was failing the hash was never written and every build re-ran it. Once the install succeeds the hash gets written and subsequent builds print `package.json has not been modified.` and skip it.

Details, empirical verification, and the narrower residual issue (the first build after a fresh install still pays one redundant install) are in #57 and in a correction comment on #54.

</details>

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -518,12 +518,21 @@ const ENGINE_RETRY_NOTICE = '\n⚠ This site requires a newer Node.js than this 
 // relaxed. The first failure's output still reaches the log, so the real reason
 // stays visible instead of being silently papered over.
 //
-// The automatic retry is opt-in and only npm:install enables it. Scripts
-// (build, grunt, …) can fail partway through with side effects already on
-// disk, and npm prints EBADENGINE as a mere warning when engine-strict is off
-// — so a warning followed by an unrelated non-zero exit would wrongly restart
-// a half-finished script. An install, by contrast, is idempotent to re-run.
-function runNpmWithEngineRetry({ runnerPath, args, cwd, onLog, onDone, register, retryOnEngineMismatch = false }) {
+// Two independent knobs, and each command uses exactly one of them:
+//
+// `retryOnEngineMismatch` (npm:install) runs strict first and retries relaxed.
+// Scripts (build, grunt, …) must not do this: they can fail partway through
+// with side effects already on disk, and npm prints EBADENGINE as a mere
+// warning when engine-strict is off — so a warning followed by an unrelated
+// non-zero exit would wrongly restart a half-finished script. An install, by
+// contrast, is idempotent to re-run.
+//
+// `relaxEnginesFromStart` (npm:run-script) is the opposite trade and is safe
+// for exactly that reason: nothing is ever restarted, engine checks are simply
+// off from the first process onward. Scripts need it because they spawn nested
+// installs that inherit this environment — wordpress-develop's Gruntfile calls
+// install-changed at load time, which execSync's its own `npm install` (#54).
+function runNpmWithEngineRetry({ runnerPath, args, cwd, onLog, onDone, register, retryOnEngineMismatch = false, relaxEnginesFromStart = false }) {
 	const start = (relaxEngines) => {
 		const child = spawn(process.execPath, [runnerPath, ...args], {
 			cwd,
@@ -563,7 +572,7 @@ function runNpmWithEngineRetry({ runnerPath, args, cwd, onLog, onDone, register,
 			onDone(code);
 		});
 	};
-	start(false);
+	start(relaxEnginesFromStart);
 }
 
 ipcMain.handle('npm:install', async (event, directoryPath) => {
@@ -601,6 +610,10 @@ ipcMain.handle('npm:run-script', async (event, directoryPath, scriptName, script
 		runnerPath: path.join(__dirname, 'script-runner.js'),
 		args: [directoryPath, scriptName, ...scriptArgs],
 		cwd: directoryPath,
+		// Baseline-relax rather than retry — see runNpmWithEngineRetry. Without
+		// this the nested `npm install` a build task spawns fails with EBADENGINE
+		// before grunt even starts (#54).
+		relaxEnginesFromStart: true,
 		register: (child) => {
 			runningScripts[runId] = child;
 			runIdByDirectory[directoryPath] = runId;
--- a/test/fixtures/nested-install/.npmrc
+++ b/test/fixtures/nested-install/.npmrc
@@ -1,0 +1,4 @@
+# Mirrors wordpress-develop's .npmrc, which is what turns an engines mismatch
+# from a warning into a fatal error — including for the nested install the
+# build script spawns.
+engine-strict = true
--- a/test/fixtures/nested-install/needs-future-node/package.json
+++ b/test/fixtures/nested-install/needs-future-node/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "needs-future-node",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Declares a Node version that will never exist, so this fixture keeps reproducing EBADENGINE no matter how new the Node running the tests is. A real registry package would stop failing once CI's Node caught up, silently turning the test green.",
+  "engines": {
+    "node": ">=99.0.0"
+  }
+}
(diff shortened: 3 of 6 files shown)


juanmaguitar:
## Relationship with #40 (Electron 43 / Node 24)

#40 is approved and would upgrade the bundled runtime to Node 24.18.0. That removes this PR's immediate motivation: with Node 24 there is no `EBADENGINE` wall for the nested install to hit, so `relaxEnginesFromStart` would have nothing to relax. Same goes for the retry path from #38, already on `trunk`.

Keeping it anyway, deliberately. #46 predicted the recurrence that produced this PR in the first place — *"the same class of failure will recur as `wordpress-develop`'s dev dependencies advance"* — and it did, one level deeper than where it was first fixed. Baseline-relax for scripts costs nothing when engines already match, and it is what stops the next floor bump from resurfacing as unreadable npm noise in the build log. Recorded here so that nobody reading this in six months has to guess what it defends against.

**Merge order matters, for two reasons.** #40 touches `package.json` and `package-lock.json`, and so does #50 at the bottom of this stack (it adds `electron-log`) — a lockfile conflict is unavoidable whichever goes first. And #40's own description notes its Windows and macOS artifacts still need a launch and workflow smoke test, whereas this stack has been verified end to end on Windows. Landing the stack first means #40 gets rebased onto validated ground rather than the reverse.

Note that #40 does **not** make #56 redundant: Node's refusal to spawn a `.bat`/`.cmd` without `shell: true` (CVE-2024-27980) is still in place in Node 24, so `spawn EINVAL` would come straight back without it.

Whichever order is chosen, the Windows pass has to be repeated after #40 lands — it changes the runtime underneath all five fixes, so the existing validation stops covering them.

PR #62, opened 2026-07-29
Test DO NOT MERGE: Electron 43 (#40) on top of the Windows fix stack

### LINKED ISSUE #37: "Install npm dependencies" fails with EBADENGINE: bundled Node is older than wordpress-develop requires

juanmaguitar:
> **Not for merge.** This is a test vehicle: the Windows fix stack (`trunk` → #50 → #51 → #52 → #56 → #58) with #40 merged on top, so Buildkite can produce a single Windows artifact built from both. Base is the stack tip, so the diff below is only the #40 merge.

## Why

#40 upgrades Electron 32.3.3 → 43.2.0 (bundled Node 24.18.0) to fix #37. Its Buildkite builds pass, but the packaged app fails on Windows — apparently at the same points this stack already addresses: lost output and flashing consoles (#52), `spawn EINVAL` at `grunt gutenberg:verify` (#56), and `EBADENGINE` from the nested `install-changed` install (#58).

Neither branch alone answers the question. This one is meant to separate:

- **the stack cures it** → the Windows failure in #40 is pre-existing, and #40 just needs to land after the stack (or be rebased onto it);
- **it fails at the same point anyway** → the stack does not cover it and Electron 43 breaks something of its own on Windows;
- **it fails somewhere new** → Electron 43 regresses code the stack already fixed. The on-disk app log from #50 (Help → open log) is directly comparable against the same run on the stack alone.

## How it was combined

Merged `codex/electron-43-spike` into the stack tip with `--no-ff`. The two sides barely overlap — #40 touches `package.json`, `package-lock.json`, `.buildkite/*`, `.nvmrc`, `README.md`; the stack touches `src/**`, `test/**` and adds one dependency.

- `package.json` is #40's, plus the stack's `electron-log` line. Nothing else.
- `src/` and `test/` are byte-identical to the stack tip — #40 touches neither, so nothing in the five PRs is altered.
- The full file set against `trunk` is exactly the union of both sides; no third file crept in.

`package-lock.json` was the only real merge. Git's three-way resolution came out clean, which is not by itself trustworthy given #40 regenerated the whole lockfile, so it was checked independently: resetting to #40's lockfile and letting `npm install --package-lock-only` re-add `electron-log` produced a byte-identical result. It resolves electron 43.2.0, electron-builder 26.15.3, `@wp-playground/cli` 3.1.47, `fs-ext-extra-prebuilt` 2.2.7, electron-log 5.4.4.

## Validated on macOS (arm64)

- `npm ci` clean; the postinstall rebuilt `fs-ext-extra-prebuilt` against Electron 43 without complaint.
- `npm test` — 71/71, **0 skipped**. The stack's Windows tests (`win-spawn-patch`, `hide-child-windows`) simulate the platform, so they really ran here rather than being skipped off-Windows.
- `npm run build:once` passes.

None of that touches Windows. That is what this PR is opened for.

## What's still needed

The Windows artifact from Buildkite, then the usual pass on a Windows VM: clone → install → build → dev server → wizard → wp-admin — the same walkthrough used to verify the stack.


juanmaguitar:
Related PRs have been merged to trunk. So closing this PR

PR #79, opened 2026-07-31
Bind the dev server to loopback only

### LINKED ISSUE #59: Dev server binds to all network interfaces, triggering a Windows Firewall prompt and exposing the site to the LAN
### Description

Starting the dev server triggers a Windows Firewall prompt — *"Do you want to allow public and private networks to access this app?"*. It is not a false positive: the Playground server really does bind to every network interface, so the contributor's development WordPress — with its `admin`/`admin` credentials — is reachable from the local network. On Contributor Day wifi that means every attendee's dev site is exposed to the room, and the firewall prompt is the only thing standing in the way.

Expected: the server binds to `127.0.0.1` only, and no firewall prompt ever appears.

1. Create a site, install dependencies, run the first full build.
2. Click "Start dev server and finish the wizard".
3. Windows Firewall asks whether to allow public and private networks to access the app.
4. Allow it, then from another machine on the same network open `http://<vm-ip>:<port>/` — the site answers.

### Operating system

Windows (the prompt is Windows-only, but the all-interfaces binding happens on macOS and Linux too — there it is simply silent).

### App version

Windows build from Buildkite, branch `juanmaguitar/windows-fixes-integration` (PR #55), commit `bc0e163`.

### Technical notes

- `src/server-runner.js:22` calls `runCLI({ command: 'server', ... })` without any host.
- In `@wp-playground/cli@3.0.8`, the server starts with `server.listen(port, callback)` — no address argument (`run-cli-Md2TQOSk.js:23`). With no host, Node listens on `::` / `0.0.0.0`, i.e. all interfaces.
- `src/playground-web-runner.js:22` **does** pass `host: '127.0.0.1'`, but `RunCLIArgs` (`run-cli.d.ts`) has no `host` option — only `db-host`. That line is silently ignored, so it gives a false sense of safety while the web server binds everywhere too.
- The SMTP bridge is fine: `src/main.js:234` already does `server.listen(0, '127.0.0.1', resolve)`.

**Suggested fix.** Patch `net.Server.prototype.listen` in `src/server-runner.js` and `src/playground-web-runner.js` before the Playground CLI is required, defaulting the host to `127.0.0.1` when the caller passes only a port. This mirrors the pattern already used by `src/hide-child-windows.js` (#52), which is required at the top of `server-runner.js` for the same "patch before loading the CLI" reason.

Worth doing upstream as well: a `host` option on `runCLI` would remove the need for the patch and fix the dead `host` argument in `playground-web-runner.js`. That shouldn't block this, though — the local patch is what stops shipping a LAN-exposed dev site.

juanmaguitar:
Fixes #59.

## The problem

`@wp-playground/cli` starts its server with `server.listen(port, callback)` — no address — so Node binds to `::`/`0.0.0.0`, every interface. The contributor's WordPress, `admin`/`admin` credentials and all, is served to the local network. On Windows that is what raises the firewall prompt at every dev-server start; on macOS and Linux it happens silently.

`runCLI()` has no `host` option to pass instead — `RunCLIArgs` only knows `db-host` — which also means the `host: '127.0.0.1'` that `playground-web-runner.js` was passing had never done anything. It just read as safe.

## The fix

`src/bind-loopback.js` patches `net.Server.prototype.listen` to fill in `127.0.0.1` when the caller gives a port and no address, and both runners apply it before requiring the Playground CLI. Same shape as `hide-child-windows.js` (#52), and for the same reason: the CLI can't be told, so the built-in is patched before it loads. `http.Server` extends `net.Server` without overriding `listen`, so one patch covers the HTTP server.

It's a default, not a lock — an explicit address (including a deliberate `0.0.0.0`), an IPC path and a handle all pass through untouched. The patch lives only in the two runner subprocesses; the main process is unchanged, and its SMTP bridge was already loopback-only.

The dead `host` argument in `playground-web-runner.js` is removed, with a comment pointing at where the protection actually comes from now.

## Verification

`node --test test/bind-loopback.test.cjs` — 11 tests over every `listen` overload. The rest of the suite is unaffected.

Against a real server, before and after the patch:

```
unpatched: {"address":"::","family":"IPv6","port":49697}
patched:   {"address":"127.0.0.1","family":"IPv4","port":49698}
```

**Still needs a check on a real machine.** On Windows, start the dev server on a box that hasn't previously allowed the app through the firewall — no prompt should appear, and `netstat -ano | findstr :<port>` should show `127.0.0.1:<port>`. On macOS/Linux, `lsof -nP -iTCP -sTCP:LISTEN | grep <port>` should show `127.0.0.1:<port>` rather than `*:<port>`, and `curl http://<lan-ip>:<port>/` should refuse while `127.0.0.1` answers. Worth exercising both the dev server and the Playground web runner.

## Upstream

Filed as WordPress/wordpress-playground#4224 — a loopback default and a real `host` option on `runCLI`. When that lands, `src/bind-loopback.js` and the workaround both go away. Not a blocker: the patch is what stops shipping a LAN-exposed dev site today.

### DIFF
--- a/src/bind-loopback.js
+++ b/src/bind-loopback.js
@@ -1,0 +1,87 @@
+// Keeps the Playground servers on the loopback interface.
+//
+// `@wp-playground/cli` starts its server with `server.listen(port, callback)` —
+// no address — and Node then listens on `::`/`0.0.0.0`, i.e. every interface.
+// That makes the contributor's WordPress, admin/admin credentials and all,
+// reachable from the local network, and on Windows it is what raises the
+// firewall prompt asking to allow public and private networks.
+//
+// `runCLI()` has no `host` option to pass instead (`RunCLIArgs` only knows
+// `db-host`), so the only place to fix it from the outside is `listen` itself.
+// Patching `net.Server.prototype.listen` before the CLI is required covers every
+// server it creates — `http.Server` extends `net.Server` and doesn't override
+// `listen`. Same "patch the built-in first" shape as `hide-child-windows.js`.
+//
+// Delete this once `runCLI` accepts a host.
+
+const PATCHED = Symbol.for('wp-dev-env.loopbackPatched');
+
+const LOOPBACK = '127.0.0.1';
+
+function isPlainObject(value) {
+	return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+// Node treats a numeric string port the same as a number.
+function isPort(value) {
+	if (typeof value === 'number') return Number.isInteger(value) && value >= 0;
+	return typeof value === 'string' && /^\d+$/.test(value);
+}
+
+// Given the arguments of `net.Server.prototype.listen`, returns them with a
+// loopback host filled in — but only when the caller asked for a TCP port and
+// left the address out. Every other overload is passed through untouched:
+// `listen(path, ...)` is an IPC socket, `listen(handle, ...)` carries its own
+// binding, and an explicit host (even a deliberate '0.0.0.0') wins.
+function withLoopbackHost(args, host = LOOPBACK) {
+	if (args.length === 0) return args;
+
+	const [first] = args;
+
+	if (isPlainObject(first)) {
+		// Options form. `path` means IPC; anything already bound to a host or
+		// carrying its own handle is left alone.
+		if (!isPort(first.port)) return args;
+		if (first.host !== undefined || first.path !== undefined) return args;
+		const next = args.slice();
+		next[0] = { ...first, host };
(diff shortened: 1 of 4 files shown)


juanmaguitar:
Verified manually on Windows after merge: the dev server behaves as expected, no firewall prompt and the site is not reachable from the LAN.

Noting it here because the workaround in `src/bind-loopback.js` is meant to be temporary — when WordPress/wordpress-playground#4224 lands a real `host` option on `runCLI`, whoever removes this patch should know the loopback behavior was confirmed on a real machine, not just in unit tests.

PR #96, opened 2026-07-31
Do not mark a site initialized when npm install fails

### LINKED ISSUE #42: Initialization wizard treats partial npm installs as complete and prevents recovery
## Summary

The initialization wizard treats any non-empty `node_modules/` directory as a successful dependency installation, even when `npm install` failed or was interrupted. It also marks a site initialized regardless of the install process's exit code.

That false-positive state unlocks **Run first full build**, disables the install button, and sends the user into a downstream failure such as `grunt: command not found` without offering a recovery path.

## Environment where reproduced

- macOS Darwin 25.5.0.
- App running from source.
- Electron runtime on PR #40: Node 24.18.0 and npm 10.9.3.
- Site: a current `wordpress-develop` checkout reporting `WordPress@7.1.0`.
- The same WordPress site was visible while switching between `trunk` and PR #40 because Electron Store and the on-disk site directory are shared across source branches.

## Observed failure

After the wizard considered dependencies installed and enabled **Run first full build**, manually running the command in the embedded terminal produced:

```text
npm run build
Running npm run build…
npm verbose cli /Users/ashfame/git/wordpress/experimental-wp-dev-env/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron npm
npm info using npm@10.9.3
npm info using node@v24.18.0
npm warn config production Use `--omit=dev` instead.
npm verbose title npm run build
npm verbose argv "run" "build"
npm verbose logfile logs-max:10 dir:/Users/ashfame/.npm/_logs/2026-07-27T21_23_24_663Z-
npm verbose logfile /Users/ashfame/.npm/_logs/2026-07-27T21_23_24_663Z-debug-0.log

> WordPress@7.1.0 build
> grunt build

sh: grunt: command not found
npm verbose cwd /Users/ashfame/a8c:experimental-wp-dev-env/first
npm verbose os Darwin 25.5.0
npm verbose node v24.18.0
npm verbose npm  v10.9.3
npm verbose exit 127
npm verbose code 127
npm run build exited with code 127
```

Current `wordpress-develop` declares `grunt` in `devDependencies` and maps its `build` script to `grunt build`: https://github.com/WordPress/wordpress-develop/blob/trunk/package.json

Therefore, `grunt: command not found` means the site's dependency installation was incomplete or invalid even though the toolkit presented it as complete.

## Steps to reproduce

One deterministic way to reproduce the state problem is:

1. Create a new site through the initialization wizard.
2. Start **Install npm dependencies**.
3. Cause npm to exit non-zero after creating some of `node_modules/`, for example by using the current `trunk` runtime that fails with `EBADENGINE` as documented in #37, interrupting installation, or reproducing another lifecycle failure.
4. Reopen or refresh the site.
5. Observe that the wizard labels dependencies installed because `node_modules/` exists and is non-empty.
6. Observe that the install button is disabled and **Run first full build** is enabled.
7. Run `npm run build`.
8. Observe a missing dependency or executable such as `grunt: command not found`.

Switching source branches while reusing the same site makes the problem easier to encounter: the failed site's directory and Electron Store state survive the switch, so a runtime fix on another branch can inherit the earlier partial installation.

### 1. Installation success is inferred from directory contents

`site:status` sets `hasNodeModules` to true whenever `node_modules/` exists and contains at least one entry: https://github.com/WordPress/experimental-wp-dev-env/blob/trunk/src/main.js#L362-L378

This cannot distinguish a successful install from npm's partially populated working directory after an error, cancellation, network failure, incompatible engine, or failed lifecycle script.

### 2. Every completed install is marked initialized, including failures

The renderer contains an explicit TODO saying not to mark failed installs initialized, but the implementation uses `if (1 || code === 0)`, making the success path unconditional: https://github.com/WordPress/experimental-wp-dev-env/blob/trunk/src/renderer/index.jsx#L829-L863

Consequently, any exit code invokes `markSiteInitialized()` and `onInitialized()`.

### 3. The UI disables the recovery action

The install checklist button is disabled whenever `hasNodeModules` is true: https://github.com/WordPress/experimental-wp-dev-env/blob/trunk/src/renderer/index.jsx#L1376-L1390

Once a partial directory is misclassified, users cannot retry installation from the checklist even though the required dependencies are missing.

### 4. The build readiness gate trusts the same weak signal

The build step uses `hasNodeModules` as its readiness condition, so a partial install immediately unlocks the next operation: https://github.com/WordPress/experimental-wp-dev-env/blob/trunk/src/renderer/index.jsx#L1392-L1405

## Expected behavior

- A non-zero npm exit must leave the dependency step failed or incomplete.
- The site must not be marked initialized after a failed installation.
- The UI must continue to offer **Retry npm install**.
- A partial `node_modules/` directory must not unlock the build step.
- The actual npm error and exit code must remain visible.
- Returning to the site or switching app versions must preserve an accurate incomplete state.

## Suggested implementation direction

Persist an explicit successful-install state only after the child exits with code 0. Filesystem inspection can supplement that state, but a non-empty directory should not be the primary success signal.

For imported/pre-existing sites, validate a meaningful installation invariant rather than directory non-emptiness. Possibilities include resolving required executables such as `node_modules/.bin/grunt`, checking the installed root package, or providing an explicit **Use existing installation** validation path. This should remain compatible with the import/offline work proposed in #12.

On failure, keep or relabel the button as **Retry npm install**. A retry can allow npm to repair the existing directory; destructive cleanup should be an explicit separate action rather than silently deleting user data.

Remove the unconditional `1 ||` success path and make renderer state reflect the real exit code.

## Regression coverage

- Simulate an install that creates a non-empty `node_modules/` and exits non-zero; status must remain incomplete.
- Verify that a failed install does not call `markSiteInitialized()`.
- Verify that the install/retry button remains enabled after failure.
- Verify that the build action remains locked after failure.
- Simulate a successful install and verify that the dependency step completes and build becomes available.
- Reopen a site after both scenarios and verify that persisted state remains correct.

## Acceptance criteria

- Non-zero installation exits cannot produce a completed dependency step.
- A partial `node_modules/` directory cannot by itself produce `hasNodeModules: true` for wizard purposes.
- Failed installations expose a retry action.
- Build remains locked until dependency installation is verified.
- A successful retry repairs the state without requiring the user to delete and recreate the entire site.
- Existing valid/imported sites have a documented validation path.

## Related

- #37 documents the current `EBADENGINE` failure that can create or expose this state.
- #12 proposes importing pre-built sites and relies on status detection, so its implementation should not institutionalize the same non-empty-directory heuristic.
- PR #40 fixes the underlying Node engine mismatch but cannot repair a partial site inherited from an earlier failed run.
- #41 tracks the invisible output from the **Run first full build** checklist button.

juanmaguitar:
## Why

The wizard's install-done callback still contained a debug short-circuit — `if (1 || code === 0)` with its own `@TODO` — left over from diagnosing a Windows install failure. Any failed "Install npm dependencies" run marked the site initialized anyway, so a newcomer with a broken install advanced to a build step that could not succeed, with no route back (#42).

The Windows failure the short-circuit papered over (EBADENGINE on the bundled Node) has since been fixed properly (#37/#46/#54), so the workaround now only hides real failures.

## What

Gate `markSiteInitialized` on a zero exit code and replace the pasted stack-trace comment with the rationale. On failure the step stays incomplete and the install button remains available for a retry; the exit code is already printed to the terminal.

`src/renderer/index.js` is the committed esbuild output for the same change.

## Testing

`npm test` — 100 pass. Manual check is behavioral: kill the network mid-install and the site must remain "Not initialized" with the install step still actionable.

Fixes #42

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -415,9 +415,9 @@ ipcMain.handle('site:status', async (_e, sitePath) => {
 		const meta = s.get('siteMeta') || {};
 		const m = meta[sitePath] || {};
 
-		return { hasNodeModules, hasBuilt, skipInitWizard: Boolean(m.skipInitWizard), initialized: Boolean(m.initialized) };
+		return { hasNodeModules, hasBuilt, skipInitWizard: Boolean(m.skipInitWizard), initialized: Boolean(m.initialized), installFailed: Boolean(m.installFailed) };
 	} catch (e) {
-		return { hasNodeModules: false, hasBuilt: false, skipInitWizard: false, initialized: false };
+		return { hasNodeModules: false, hasBuilt: false, skipInitWizard: false, initialized: false, installFailed: false };
 	}
 });
 
@@ -678,7 +678,19 @@ ipcMain.handle('npm:install', async (event, directoryPath) => {
 		onLog: (type, data) => {
 			event.sender.send('npm:install:log', { installId, type, data });
 		},
-		onDone: (code) => {
+		onDone: async (code) => {
+			// Recorded before the done event so the renderer's status reload
+			// sees the outcome. node_modules existing is not evidence the
+			// install succeeded (#42); this flag is what site:status reports
+			// so a failed install's partial node_modules doesn't read as a
+			// completed setup step. Sites that predate the flag read as not
+			// failed, which matches the old behaviour.
+			try {
+				const s = await getStore();
+				const meta = s.get('siteMeta') || {};
+				meta[directoryPath] = { ...(meta[directoryPath] || {}), installFailed: code !== 0 };
+				s.set('siteMeta', meta);
+			} catch {}
 			event.sender.send('npm:install:done', { installId, code });
 			delete runningInstalls[installId];
 		}
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -54206,7 +54206,7 @@ If there's a particular need for this, please submit a feature request at https:
 install exited with code ${code}
 `);
         setInstalling(false);
-        if (1) {
+        if (code === 0) {
           try {
             await window.api.markSiteInitialized(sitePath);
           } catch {
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -30976,20 +30976,22 @@ WARNING: This link could potentially be dangerous`)) {
         const installing = Boolean(flags.installing);
         const building = Boolean(flags.building);
         const starting = Boolean(flags.starting);
+        const installFailed = Boolean(flags.installFailed);
+        const installOk = hasNodeModules && !installFailed;
         return {
           download: {
             done: !isPending,
             ready: true
           },
           install: {
-            done: hasNodeModules,
+            done: installOk,
             ready: !isPending,
-            disabled: isPending || statusLoading || installing || hasNodeModules
+            disabled: isPending || statusLoading || installing || installOk
           },
           build: {
             done: hasBuilt,
-            ready: hasNodeModules,
-            disabled: statusLoading || building || !hasNodeModules || hasBuilt
+            ready: installOk,
+            disabled: statusLoading || building || !installOk || hasBuilt
           },
           dev: {
             done: false,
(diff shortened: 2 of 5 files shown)


juanmaguitar:
Windows VM testing of the first commit caught a second, independent completion signal: the checklist derives the install step's done state from `node_modules` existing, and a failed install leaves a partial one behind — so the step still showed COMPLETED, its button was disabled, and the build unlocked (the "prevents recovery" half of #42), even though the initialized badge now correctly stayed off.

Second commit records each install run's outcome in the site's metadata (written before the done event), reports it from `site:status`, and lets it override the existence check in `computeSetupStepState`. After a failed install the step stays incomplete and the button reads "Retry npm install"; sites that predate the flag are unaffected. Two new tests in `test/setup-steps.test.cjs`; the first fails against the previous behavior.

Issue #84, opened 2026-07-31
Ctrl+C cannot cancel "Install npm dependencies"


juanmaguitar:
## What happens

Ctrl+C in the in-app terminal during "Install npm dependencies" does nothing: the install keeps running, the terminal's running state stays on, and input stays frozen until the install finishes on its own. This is the longest operation in the app and the one a user most needs to abort (wrong network, corporate proxy, changed their mind).

## Root cause

The kill request handler (`src/main.js:724-741`) searches only the script-run registries, which are populated exclusively by the run-script path. The install registry (`src/main.js:122`, written at `:676`) is never read by anything — installs are tracked but unkillable. The renderer's Ctrl+C handler (`src/renderer/index.jsx:1028`) gets back a quiet "no running script" failure and shows nothing.

## Expected

Cancelling an install stops the process (including its children) and returns the terminal to an idle state, same as cancelling a script.

juanmaguitar:
Closing: fixed by #274, with the retry half covered by #414, and confirmed by hand on a packaged build.

**What changed.** `npm:kill` in `src/main.js` falls back to the install registry by directory when no script matches, so the request the terminal's Ctrl+C sends (`killCurrent` always passes `directoryPath`) finds the running install and ends its whole tree through `killChildTree`. The handler marks the child cancelled before killing it, so `runNpmWithEngineRetry` does not read the non-zero exit as an engine mismatch and respawn the install that was just stopped.

**What pins it.** `tests/unit/ipc-wiring.test.cjs`, "npm:kill ends a running install, which only its directory can name": starts an install, kills by directory alone, asserts the tree kill and that the close that follows starts no second child. The #414 tests cover the same stop from the delete path.

**Manual pass, 2026-09-11, Windows 11, packaged build from trunk `ac31720` (Buildkite #993).** Started "Install npm dependencies" on a Core site with `node_modules` removed, pressed Ctrl+C in the embedded terminal while npm was still downloading: the terminal printed `^C`, the output stopped within a few seconds and the prompt came back. The app did not start a second install on its own. `npm install` typed into the terminal afterwards ran to `npm install exited with code 0`.

Not established by this pass: that no child of the stopped install outlived it. The install runs as Electron under `ELECTRON_RUN_AS_NODE`, so `Get-Process node,npm` sees nothing either way, and counting the app's own processes gave the same number during the install and at rest. The unit test covers the tree kill; the by-hand evidence here is the terminal returning and no relaunch.

PR #98, opened 2026-07-31
Kill spawned process trees when the app quits

### LINKED ISSUE #83: Quitting the app leaves dev servers and watchers running
## What happens

Quitting the app while a dev server or a watch task is running leaves those processes alive. The Playground server keeps its port and the watcher keeps rebuilding until they are killed by hand — at a Contributor Day that means Task Manager / Activity Monitor. On the next launch, the web-server "already reachable" probe (`src/main.js:876-888`) can silently adopt the previous session's orphaned server.

## Root cause

Two independent gaps:

1. There is no quit hook anywhere in `src/main.js` — only the window-all-closed handler (`src/main.js:392-394`). Nothing sweeps the tracked children (`runningInstalls`, `runningScripts`, `playgroundServers`, the web server) on exit.
2. Even when the app does kill a child (`src/main.js:736, 848, 995`), the signal reaches only the direct child. The processes that matter are grandchildren (script runner → shell → grunt → node); on Windows especially, the tree survives.

Related leak: the Windows shim directory created per run in the temp folder (`src/main.js:45`) is never removed — one leftover directory per launch.

## Expected

Quitting the app ends everything it started, on all platforms.

juanmaguitar:
## Why

Quitting the app left its children running (#83), for two stacked reasons:

- No quit hook existed, so nothing swept the tracked children at all.
- Even where the app does kill a child, `child.kill()` signals only the direct runner — the interesting processes are grandchildren (runner → npm → shell → grunt → node), and on Windows especially the tree survives.

The practical result at a Contributor Day: quit with a dev server running and `grunt _watch` keeps rebuilding while the WASM PHP server keeps its port, with Task Manager as the only remedy. On relaunch, the web server's already-reachable probe can silently adopt the orphaned server.

## What

- New `src/kill-tree.js`: kills a child together with its descendants. On Windows, `taskkill /pid <pid> /T /F`; on POSIX, a SIGTERM to the process group with a bare-pid fallback. The platform decision is a pure function (`killTreePlan`) so the dangerous part is testable without spawning anything — including the guard that a pid of 0/-1 must never produce a plan.
- The three spawn sites (npm runner, per-site server, web server) now pass `detached: true` on POSIX so each child leads a process group its descendants stay inside. No behavior change on Windows, and no change to stdio or lifetime otherwise.
- A `before-quit` handler sweeps every registry of running children (installs, scripts, per-site servers, web server) through `killChildTree`.

Explicitly out of scope: making Ctrl+C able to cancel an install (#84) and the mid-session kill paths — this PR only guarantees quit cleans up.

## Testing

`npm test` and `npm run test:electron` — 107 pass each (7 new in `test/kill-tree.test.cjs`). Manual check: start the dev server, quit the app, and verify no `grunt`/PHP server processes remain and the port is free on relaunch — on both macOS and Windows.

Fixes #83

### DIFF
--- a/src/kill-tree.js
+++ b/src/kill-tree.js
@@ -1,0 +1,60 @@
+// Kills a spawned child together with its descendants.
+//
+// The processes worth stopping are rarely the direct child: the runners spawn
+// npm, which spawns a shell, which spawns grunt, which spawns node. A plain
+// child.kill() signals only the first link, so quitting the app left watchers
+// and servers running (#83).
+//
+// Two platform mechanisms, chosen by killTreePlan so the decision is testable
+// without spawning anything:
+//
+// - win32: `taskkill /pid <pid> /T /F` — the only stock way to end a process
+//   tree on Windows, where signals do not exist and job objects are not
+//   available to an already-spawned tree.
+// - POSIX: signal the process group (negative pid). This requires the child to
+//   have been spawned with `detached: true`, which makes it a group leader its
+//   descendants stay inside; falls back to signalling the child alone when the
+//   group signal fails (e.g. the child was not detached).
+
+'use strict';
+
+/**
+ * Pure decision: how to kill the tree rooted at `pid` on `platform`.
+ * Returns null for a pid that cannot identify a live process.
+ */
+function killTreePlan(platform, pid) {
+	if (!Number.isInteger(pid) || pid <= 0) return null;
+	if (platform === 'win32') {
+		return { type: 'command', command: 'taskkill', args: ['/pid', String(pid), '/T', '/F'] };
+	}
+	return { type: 'signal', signal: 'SIGTERM', target: -pid, fallback: pid };
+}
+
+/**
+ * Applies killTreePlan to a ChildProcess. Never throws: this runs during
+ * quit, where a failure to kill one child must not stop the sweep of the rest.
+ * Returns true when a kill was attempted.
+ */
+function killChildTree(child, {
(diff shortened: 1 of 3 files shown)


juanmaguitar:
Windows VM testing caught a survivor: after quitting with the dev server running, a `grunt _watch` grandchild stayed alive (`ParentProcessId` pointing at an already-dead intermediate). That's the documented weakness of `taskkill /T` — it walks parent links at kill time, so a dead intermediate orphans everything below it.

Second commit snapshots the process table (`Get-CimInstance Win32_Process`) while the chains are intact, computes the descendant closure of the tracked roots, and kills each pid individually, keeping `taskkill /T` as backstop and fallback. The closure and the sweep dispatch are pure and covered by 5 new tests, including the exact observed leak shape (root → dead intermediate → surviving grandchild).

juanmaguitar:
Simplified back to the plain sweep after discussion: the snapshot-and-closure approach added quit-time machinery that didn't survive VM testing anyway. This PR now ships `taskkill /T` + POSIX group-kill + the `before-quit` sweep — complete cleanup on macOS/Linux and whenever the Windows tree is intact.

Known residual gap, kept deliberately and documented in the code: on Windows a grandchild whose intermediate parent died first can survive the sweep (observed once with `grunt _watch` on a VM). That's the same class of process trunk leaks on *every* quit today, so this is still a strict improvement. Changed "Fixes #83" to "Part of #83" — the issue stays open for the complete Windows tree-kill.

PR #99, opened 2026-07-31
Show the folder name, not the full path, for Windows sites

### LINKED ISSUE #87: Windows: sites without a label show their full path in the sidebar
## What happens

On Windows, a site that has no stored label (any site registered by pointing at an existing directory rather than created through the wizard) shows its **entire path** in the sidebar and in the site heading — `C:\Users\me\wp-sites\my-site` instead of `my-site`.

## Root cause

The renderer derives the display name by splitting the site path on forward slashes, in two places (`src/renderer/index.jsx:505` and `:746`). Windows paths use backslashes, so the split returns the whole path unchanged. The main process handles this correctly with the platform-aware basename; the renderer duplicates the logic with a hardcoded separator.

## Expected

The last path segment on every platform, from one shared helper instead of two copies.

juanmaguitar:
## Why

The renderer derived a site's display name by splitting the site path on forward slashes, in two independent places (sidebar entry and site heading). Windows paths use backslashes, so any site without a stored label — i.e. one registered by pointing at an existing directory rather than created through the wizard — showed its entire `C:\Users\...` path (#87). The main process already handles this correctly via `path.basename`; the renderer had duplicated the logic with a hardcoded separator.

## What

A new `src/renderer/path-basename.cjs` (the renderer has no Node `path` module) that takes the last segment across `/`, `\`, and mixed separators, used from both display sites. Follows the repo's extracted-tested-module pattern with `test/path-basename.test.cjs`; the Windows cases in that test fail against the old `split('/')` behavior.

`src/renderer/index.js` is the committed esbuild output for the same change.

## Testing

`npm test` — 104 pass (4 new). Manual check on Windows: add an existing folder as a site without renaming it — the sidebar and heading must show the folder name only.

Fixes #87

### DIFF
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -31035,6 +31035,18 @@ WARNING: This link could potentially be dangerous`)) {
     }
   });
 
+  // src/renderer/path-basename.cjs
+  var require_path_basename = __commonJS({
+    "src/renderer/path-basename.cjs"(exports, module) {
+      "use strict";
+      function pathBasename2(p) {
+        const s = String(p ?? "");
+        return s.split(/[\\/]/).filter(Boolean).pop() || s;
+      }
+      module.exports = { pathBasename: pathBasename2 };
+    }
+  });
+
   // src/renderer/index.jsx
   var import_react69 = __toESM(require_react());
   var import_client2 = __toESM(require_client());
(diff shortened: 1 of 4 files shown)


juanmaguitar:
This PR is now the top of a stack (#96 → #97 → #98 → #99), so the Buildkite artifact built from this branch exercises all four pre-release fixes together.

juanmaguitar:
Stack updated: head is now `4b5ea66` (adds the wizard step-state fix to #96 after Windows VM testing showed the checklist advancing on a failed install via the node_modules existence check). Test with the Buildkite artifact for this commit — earlier artifacts (`d3aefdf` and before) don't include it.

juanmaguitar:
Stack updated again: head is now `8b47653` — includes the Windows quit-sweep rework in #98 after VM testing found a surviving grunt _watch grandchild. Re-test the quit scenario (#83) with the Buildkite artifact for this commit.

juanmaguitar:
Stack updated: head is now `3353674` (simplified #98 back to the plain sweep). This is the release-candidate stack — re-test quit cleanup knowing the single grunt _watch survivor is the documented residual gap, not a failure.

Issue #257, opened 2026-08-11
The current setup step reads "In progress" before its action starts


juanmaguitar:
The initial setup checklist marks the current step — the next one a contributor should act on — with the status **"In progress"**. But that label appears the moment a step becomes the current one: before the contributor has clicked anything and before any work is happening.

On the first screen of a fresh site, *Install npm dependencies* reads as "In progress" while nothing is installing. It's misleading — it suggests the app is already busy, so a contributor may sit and wait instead of clicking, or wonder why the button is still there offering to start what looks already started.

### Expected

While a step is the next actionable one but nothing is running yet, its status should read as something like **"Ready"** (or "Next" / "To do"). **"In progress" should appear only once the step's action is actually running** — while the install, build, or dev-server start is in flight.

### Root cause

The checklist's "current" status is derived purely from *first ready, not-done step*, and that status always maps to the "In progress" label. It never consults the running flags (installing / building / starting) that already gate the step buttons. So "current" and "running" are conflated. (`src/renderer/index.jsx`, the checklist status/label derivation.)

### Note

Surfaced by the next-action cue (#252): now that the current step is ringed, the eye lands straight on the wrong label. The mislabel itself predates that work.

PR #277, opened 2026-08-11
Add a design-token layer and convert the site header onto it

ryanwelcher:
> **First of a stack.** This is Phase 1 of a seven-phase UI pass. Phases 2–7 build on it, one PR each: first-run path, setup checklist, notices, Trac card, patch/terminal panes, sidebar and cleanup.

## Why

Controls float. "Start dev server" and "Review & submit changes" sat in open space between the header and the Trac card, belonging to nothing.

That is architectural, not cosmetic. `index.jsx` is styled by 317 inline `style={{}}` objects, so every panel picks its own values: eight `marginTop`s, three border greys, two muted greys, three radii.

## Before and after

The path, its copy button and "Open directory in" were three stacked rows. They are now one meta block. The two buttons sit in a bounded row under a hairline.

**Before**

**After**

### The content column, on a wide window

The only shot taken at 1600px. Every other image uses 1200px, where the content area is ~855px — narrower than the 880px cap, so the cap has no visible effect and no screenshot could show it.

### Setup wizard

The checklist below the header is untouched — that is Phase 3.

**Before**

**After**

## What changes

**`src/renderer/styles/tokens.css`** — one place for spacing, colour, radius and type. Imported after the `@wordpress/components` stylesheet; esbuild concatenates CSS imports in order, so the cascade is right with no build change. Values follow WPDS.

**`src/renderer/ui/`** — `Section`, `ActionRow`, `MetaText`, `StatusBadge`. Each replaces something the file already did several ways.

**The site header**, converted onto both.

Three decisions worth naming:

- The dev-server button drops its 15px text and 12px radius for the stock WPDS shape. `.next-action-cue` then reads `--wpct-radius-control`, because a 10px glow around a 2px button looked like a rounded box behind a square one.
- The cue's colour becomes `--wpct-cue` — pixel-identical to the `#f0b849` it replaces, and deliberately not the warning amber. The cue means "do this next", and it is often drawn around a warning notice.
- `status-tone.cjs` maps status to tone only. `setupStepLabel` owns the checklist words and distinguishes "Ready" from "In progress" (#257); restating them here would be a second source of truth.

**Not in this PR:** the checklist, notices, Trac card, patch and terminal panes, sidebar. Two header notices keep their inline styling for the same reason — converting one now leaves two notice styles on screen until Phase 4.

**Also here, and not UI polish:** the screenshot harness now produces the same images twice (c0f8a00). Three PNGs used to change on every run with no code change — CSS animations caught mid-frame, and the linked-pull-request lookup photographed mid-flight. It is in this PR because this branch already edits both harness files and eleven of the screenshots, so a separate PR would collide with it on rebase, and a binary conflict has no resolution but to pick a side and re-capture — the exact failure the commit prevents. It matters now because six more PRs stack on this branch. The commit message has the full reasoning; #326 covers what is left.

## How to test this

**Platforms:** any (renderer-only).

**Starting state:**

1. An initialized site with the wizard skipped, so the header and action row both render.
2. A stale site, to check the amber dot and the divider under a notice.

**What must not have happened:**

- Rename, copy-path and "Open directory in" all still work. Open the menu twice — it detects applications on each open.
- The header `h1` is still the page's only `h1`. The rename pencil sits beside the heading, not inside it.
- The next-action glow still lands on one target and still scrolls it into view.
- On a maximised window the content column stays centred and stops at 880px.

## Risks and limitations

- The content column was already capped at an inline `maxWidth: 1040`; this moves it to a token and tightens it to 880px.
- **`/self-review` found 4, fixed 3.** See the review-outcome section below.
- Inline styles: 317 → 292.

## Related

Part of the UI polish pass, phase 1 of 7. Raised #278 and #326 while working on this.

---

<details>

**Rules, not boxes.** One card per section, grouped by hairlines and a spacing scale — not nested inset panels. The Trac card already has three subsections; boxes inside boxes is what would read as cluttered.

**Spacing carries meaning.** 8px within a group, 16px between groups in a card, 24px between cards. These are currently interchangeable, which is why nothing reads as grouped.

**The token table lands whole**, rather than growing per phase. It is a constants file, and re-editing it six times would churn more than it saves. Every primitive still arrives with the first phase that needs it.

**The content column is capped now.** Doing it later would re-diff every screenshot taken before it.

**The screenshot harness gained a per-shot window size** rather than a second global one. Layout that only appears past a breakpoint is invisible to a harness with one window, which is how the cap went unphotographed. Bounds are set before every shot, not only the ones that override, so a widened window cannot leak into the next image.

**Dark mode is out of scope**, and `test/color-scheme.test.cjs` holds that line. Every surface is hand-painted light and `@wordpress/components` ships light styles only.

</details>

<details>
<summary>Review outcome</summary>

Two passes. **Round 1: 3 [fix here] · 1 [follow-up] — all 3 fixed** in fed86fc. **Round 2: 1 [fix here] · 3 [follow-up] — all 4 fixed** in fcc2940.

- 🟡 architecture — a sentence laid out as a flex row, gapping every word fragment. `MetaText` gains a `flow` prop.
- 🔵 accessibility — the path chip at 4.44:1, under AA. `.wpct-meta code` also styled words it shouldn't have; it becomes an explicit `.wpct-chip` and its text moves to `--wpct-text`.
- 🔵 tests — `statusTone` read through `Object.prototype`, so `'constructor'` and `'__proto__'` returned a non-tone. Now `Object.hasOwn`; both new cases fail on the old lookup.

Deferred: `StatusBadge` renders an empty pill for a status whose label is deliberately blank. Unreachable here — only `initialized`/`uninitialized` are used, both labelled — and it is the API Phase 3's checklist conversion will call, so it is decided there.

**Round 2** re-reviewed the two commits the first pass never saw — the screenshot-harness change most of all, since six more PRs will build on it.

- 🔵 architecture — a shot's `window` was substituted for the default, not merged. `setBounds` takes a partial rectangle, so `{ width: 1600 }` with no height would inherit the previous shot's — the exact bug the call was added to prevent, as a height instead of a width.
- 🔵 architecture — `window` was undocumented in the entry-shape contract and silently ignored in the live tier. Now documented, and a live shot that declares one fails loudly.
- 🔵 tests — the wide screenshot was asserted with `existsSync`, which could never fail again once committed. It now reads the PNG's IHDR width against the width the shot declares.
- 🔵 tests — the round-1 CSS fixes shipped with no regression test. `test/meta-text.test.cjs` computes the chip's contrast from the tokens rather than hard-coding it.

Every assertion added across both rounds was checked against a mutation that should break it, because an assertion that has never failed has not been tested.

Findings 1 and 2 of round 1 were introduced by this PR, not inherited.

</details>

<details>
<summary>Implementation notes</summary>

`.next-action-cue` lives in `index.html`'s `<style>` block, emitted before the `<link>` to the bundled `index.css` that defines `:root`. That is fine — custom properties resolve at computed-value time. Verified rather than assumed: a decoded pixel diff of `site-view.png` across the colour change shows 64 differing subpixels, all in rows 411–422, which is the animated "Checking GitHub…" spinner.

`test/content-column.test.cjs` pins the cap, the class still being applied, the absence of a competing inline `maxWidth`, and the wide shot still existing and still being wide enough to prove anything. Each assertion was checked against a mutation that should break it — a widened token, a deleted rule, a renamed shot, a removed window override — because an assertion that has never failed has not been tested.

Rebased onto trunk, which moved nine commits during this work. Three reconciliations: "Forget this site" was removed from the header menu (#266); the `cueProps` wrappers and the wp-admin link (#250) were preserved inside the new `ActionRow`; and the step labels moved to `setupStepLabel` (#257), which is why `status-tone.cjs` carries no checklist words.

</details>


ryanwelcher:
> @ryanwelcher thank you for this PR. You'll notice I just [merged trunk into your branch](https://github.com/WordPress/contributor-toolkit/pull/277/commits/412fc12750993c20dc217788467000e69a0fbb4a). It's to sync with a few structural and security changes that landed on trunk since you opened this. Cheers.

Sounds great! Let me know if there is anything I can do and feel free to close this as needed.

PR #396, opened 2026-08-21
Require the safe setup for rotated macOS signing credentials

iangmaia:
## Why

We've rotated the macOS signing credentials, but branches that predate the secure signing setup (or any new branch accidentally created from an old revision) could consume them accidentally. This PR creates a new "credential epoch" by renaming the CI variables so only code with the safer setup can use the rotated credentials.

## What changes

Buildkite must supply three repository-specific `WPCT_MACOS_SIGNING_*` inputs, and setup fails with an actionable error if any is missing.

The setup script materializes the private key in a protected temporary file and exports the `APPLE_API_*` interface required by Electron Builder. Fastlane explicitly reuses that same interface. CI no longer injects the legacy names, and repository code no longer references them, so older code cannot use the rotated credentials.

## How to test this

**Starting state:** Check out this branch on macOS. The protected signing credentials are not needed for the focused unit tests.

1. Run `node --test tests/unit/macos-signing.test.cjs`. All nine tests pass, including one rejection case for each missing repository-specific variable, trace-safety and Fastlane-failure checks, and temporary-key cleanup coverage.
2. After the new variables are enabled in Buildkite, open the macOS build for the current head. `npm run dist` and `bundle exec fastlane verify_code_signing` both pass.

**What must not have happened:** Legacy variables alone must not reach Fastlane, signing material must not appear in the checkout or packaged payload, the private key must not appear in shell traces, and the temporary key must not survive the Buildkite shell.

## Risks and limitations

The new variables live outside this repository and must be configured in Buildkite before the macOS build can pass. Real signing and notarization can only be verified through Buildkite.

## Related

Follow-up to #390.

---

<details>

The repository-specific `WPCT_MACOS_SIGNING_*` names create the credential epoch without needing a version suffix, and routine future rotations can replace their values without another code change. `WPCT_MACOS_SIGNING_PRIVATE_KEY` also matches Buildkite's default `*_PRIVATE_KEY` redaction pattern. Electron Builder requires `APPLE_API_KEY` to be a filepath, so the key is materialized once and Fastlane explicitly reuses the same temporary-file-backed interface.

</details>

<details>

Final self-review: 0 `[fix here]` · 1 `[follow-up]`. Direct Fastlane invocation still accepts empty environment entries and remains a pre-existing follow-up; the normal Buildkite path rejects them. Lint, all 1,051 tests, all nine focused signing tests, and shell and Ruby syntax checks pass.

</details>

### DIFF
--- a/.buildkite/commands/setup_macos_code_signing.sh
+++ b/.buildkite/commands/setup_macos_code_signing.sh
@@ -2,20 +2,20 @@
 
 echo "~~~ :apple: Configure macOS code signing"
 
-echo "--- :ruby: Install gems"
-install_gems
-
-# The Buildkite step sources this script, so the shebang's `-eu` never applies. Return explicitly
-# so callers without `errexit` still receive materialization failures instead of continuing.
+# The Buildkite step sources this script, so the shebang's `-eu` never applies. Validate before
+# installing gems, and avoid expanding the private-key value into a command that xtrace would log.
 for required_var in \
-	WPCT_MACOS_SIGNING_KEY_ID_V2 \
-	WPCT_MACOS_SIGNING_ISSUER_ID_V2 \
-	WPCT_MACOS_SIGNING_PRIVATE_KEY_V2; do
-	if [ -z "${!required_var:-}" ]; then
-		echo "$required_var is required. Configure the V2 macOS signing credentials in Buildkite." >&2
+	WPCT_MACOS_SIGNING_KEY_ID \
+	WPCT_MACOS_SIGNING_ISSUER_ID \
+	WPCT_MACOS_SIGNING_PRIVATE_KEY; do
+	if ! printenv "$required_var" | grep -q .; then
+		echo "$required_var is required. Configure the macOS signing credentials in Buildkite." >&2
 		return 1
 	fi
 done
+
+echo "--- :ruby: Install gems"
+install_gems
 
 # Materialize the Buildkite credential once, then expose the APPLE_API_* interface used by both consumers:
 # Fastlane reads it explicitly in Fastfile, and electron-builder inherits it during the later `npm run dist`.
@@ -23,8 +23,8 @@ done
 # See https://www.electron.build/mac#notarize
 MACOS_NOTARIZATION_TEMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/wordpress-contributor-toolkit-signing.XXXXXX")" || return 1
 export APPLE_API_KEY="$MACOS_NOTARIZATION_TEMP_DIR/apple_api_key"
-export APPLE_API_KEY_ID="$WPCT_MACOS_SIGNING_KEY_ID_V2"
-export APPLE_API_ISSUER="$WPCT_MACOS_SIGNING_ISSUER_ID_V2"
+export APPLE_API_KEY_ID="$WPCT_MACOS_SIGNING_KEY_ID"
+export APPLE_API_ISSUER="$WPCT_MACOS_SIGNING_ISSUER_ID"
 
 cleanup_macos_notarization_key() {
 	rm -rf "$MACOS_NOTARIZATION_TEMP_DIR"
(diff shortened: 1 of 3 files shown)


tests/unit/macos-signing.test.cjs
```diff
@@ -71,14 +84,39 @@ exit 1
...
+		APP_STORE_CONNECT_API_KEY_ISSUER_ID: 'legacy-issuer-id',
+		SETUP_SCRIPT,
+	};
+	delete env.WPCT_MACOS_SIGNING_PRIVATE_KEY_V2;
+	delete env.WPCT_MACOS_SIGNING_KEY_ID_V2;
+	delete env.WPCT_MACOS_SIGNING_ISSUER_ID_V2;
+	return env;
+}
+
+test('legacy macOS signing variables cannot activate the rotated credentials', { skip: process.platform !== 'darwin' }, () => {
```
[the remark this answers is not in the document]
iangmaia: Fixed in f23538c.

.buildkite/commands/setup_macos_code_signing.sh
```diff
@@ -5,17 +5,22 @@ echo "~~~ :apple: Configure macOS code signing"
...
-# See https://www.electron.build/mac#notarize
-#
 # The Buildkite step sources this script, so the shebang's `-eu` never applies. Return explicitly
 # so callers without `errexit` still receive materialization failures instead of continuing.
+for required_var in \
+	WPCT_MACOS_SIGNING_KEY_ID_V2 \
+	WPCT_MACOS_SIGNING_ISSUER_ID_V2 \
+	WPCT_MACOS_SIGNING_PRIVATE_KEY_V2; do
+	if [ -z "${!required_var:-}" ]; then
+		echo "$required_var is required. This branch predates the secure macOS signing credentials; merge the latest trunk." >&2
```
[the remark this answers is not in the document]
iangmaia: Fixed in f23538c.

fastlane/Fastfile
```diff
@@ -26,7 +26,11 @@ end
 desc 'Download the development signing certificates to this machine'
 lane :setup_code_signing do |readonly: true|
   check_env_vars!
-  app_store_connect_api_key
+  app_store_connect_api_key(
+    key_id: ENV.fetch('APPLE_API_KEY_ID'),
+    issuer_id: ENV.fetch('APPLE_API_ISSUER'),
```
AliSoftware, line 31: Wouldn't it be simpler to directly reference the V2 env vars here explicitly rather that having the shell script do an `export` to export the existing V2 env vars as a new name and use the new name here? Fells like an unnecessary indirection?

fastlane/Fastfile
```diff
@@ -26,7 +26,11 @@ end
...
+    key_filepath: ENV.fetch('APPLE_API_KEY')
```
AliSoftware, line 32: Why don't we use `key_contents: ENV.fetch('WPCT_MACOS_SIGNING_PRIVATE_KEY_V2')` here instead of using `key_filepath`?!

That would completely remove the need to write the key to a temp dir on disk (and thus completely remove the risk of leaking it accidentally in the first place, and remove the need to cleanup that temp file, etc…)? Wouldn't that be not only simpler but also more secure?

AliSoftware (commented):
I think we can do even better by not only simplifying the logic a lot but also not write the private key on disk ever (even in a temp file) and use the env var with `key_contents` directly?

fastlane/Fastfile
AliSoftware: Oh I see you added the rationale in https://github.com/WordPress/contributor-toolkit/commit/8a3e1ad47815be3b11ef1aa616bbc3290203326b , thanks for documenting

fastlane/Fastfile
iangmaia: I've just pushed https://github.com/WordPress/contributor-toolkit/pull/396/changes/8a3e1ad47815be3b11ef1aa616bbc3290203326b to clarify exactly about that (also as a reply to your other comments, I think): those `APPLE_API_*` are used by electron-builder, that's why they're needed (and now reused explicitly by Fastlane).

fastlane/Fastfile
iangmaia: Yep exactly 🎯

juanmaguitar (approved):

.buildkite/commands/setup_macos_code_signing.sh
```diff
@@ -5,17 +5,26 @@ echo "~~~ :apple: Configure macOS code signing"
...
 # so callers without `errexit` still receive materialization failures instead of continuing.
+for required_var in \
+	WPCT_MACOS_SIGNING_KEY_ID_V2 \
+	WPCT_MACOS_SIGNING_ISSUER_ID_V2 \
+	WPCT_MACOS_SIGNING_PRIVATE_KEY_V2; do
+	if [ -z "${!required_var:-}" ]; then
+		echo "$required_var is required. Configure the V2 macOS signing credentials in Buildkite." >&2
+		return 1
+	fi
+done
```
AliSoftware, line 15: Given we already have `set -u` and those are provided by our IaC I don't think this is really necessary?

```suggestion
```

AliSoftware (approved):
Suggested to remove unnecessary check on setup otherwise LGTM

.buildkite/commands/setup_macos_code_signing.sh
iangmaia: I think the check is still needed because this file is `source`d, so the `#!/bin/bash -eu` flags are not applied?

AliSoftware: True that `source`-ing ignores the shebang; But for scripts that are sourced we should probably then use `set -eu` on a dedicated line at the start of the script?

An alternative that would still feel simpler than having to add a dedicated `for` loop and so many lines (especially for an error we expect will never happen now that the CI secrets are set for this repo) would be to use `${VAR?missing env var required to configure code signing}` 🤷

iangmaia: 👍 Followed-up on https://github.com/WordPress/contributor-toolkit/pull/398

PR #398, opened 2026-08-21
Simplify macOS signing failure handling

iangmaia:
## Why

Following up to the [discussion on #396](https://github.com/WordPress/contributor-toolkit/pull/396#discussion_r3830285893), this PR updates the the macOS signing setup script, which is `source`d, so its shebang's `-eu` flags are ignored. We then enable strict mode in the pipeline itself to cover the whole macOS build step.

## What changes

- The macOS Buildkite command in the .yml pipeline now enables `set -eu`.
- The signing script uses direct assertions for the key ID and issuer and keeps the private-key check safe under shell tracing and lets strict mode propagate command failures.

Tests mirror the caller behavior and cover gem installation failures.

## How to test this

- Run `npm run lint` + `npm test`.

## Risks and limitations

This affects only macOS CI. Real signing can only be exercised in Buildkite; local tests stub the external signing operations. Strict mode intentionally stops the macOS step on any command failure or unset variable.

## Related

Follow-up to #396.

---

<details>

Strict mode belongs to the caller because it governs the full command sequence. The private key still uses `printenv` because `${VAR:?…}` could expose the secret when shell tracing is enabled.

</details>

<details>

0 `[fix here]` · 0 `[follow-up]`. No findings.

</details>

### DIFF
--- a/.buildkite/commands/setup_macos_code_signing.sh
+++ b/.buildkite/commands/setup_macos_code_signing.sh
@@ -2,17 +2,14 @@
 
 echo "~~~ :apple: Configure macOS code signing"
 
-# The Buildkite step sources this script, so the shebang's `-eu` never applies. Validate before
-# installing gems, and avoid expanding the private-key value into a command that xtrace would log.
-for required_var in \
-	WPCT_MACOS_SIGNING_KEY_ID \
-	WPCT_MACOS_SIGNING_ISSUER_ID \
-	WPCT_MACOS_SIGNING_PRIVATE_KEY; do
-	if ! printenv "$required_var" | grep -q .; then
-		echo "$required_var is required. Configure the macOS signing credentials in Buildkite." >&2
-		return 1
-	fi
-done
+# The Buildkite command enables `set -eu` before sourcing this script. Validate the non-secret values
+# while exporting them, but keep the private key out of commands that xtrace would log.
+export APPLE_API_KEY_ID="${WPCT_MACOS_SIGNING_KEY_ID:?WPCT_MACOS_SIGNING_KEY_ID is required. Configure the macOS signing credentials in Buildkite.}"
+export APPLE_API_ISSUER="${WPCT_MACOS_SIGNING_ISSUER_ID:?WPCT_MACOS_SIGNING_ISSUER_ID is required. Configure the macOS signing credentials in Buildkite.}"
+if ! printenv WPCT_MACOS_SIGNING_PRIVATE_KEY | grep -q .; then
+	echo "WPCT_MACOS_SIGNING_PRIVATE_KEY is required. Configure the macOS signing credentials in Buildkite." >&2
+	return 1
+fi
 
 echo "--- :ruby: Install gems"
 install_gems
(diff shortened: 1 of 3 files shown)


.buildkite/commands/setup_macos_code_signing.sh
```diff
@@ -2,17 +2,14 @@
...
-	fi
-done
+# The Buildkite command enables `set -eu` before sourcing this script. Validate the non-secret values
+# while exporting them, but keep the private key out of commands that xtrace would log.
+export APPLE_API_KEY_ID="${WPCT_MACOS_SIGNING_KEY_ID:?WPCT_MACOS_SIGNING_KEY_ID is required. Configure the macOS signing credentials in Buildkite.}"
+export APPLE_API_ISSUER="${WPCT_MACOS_SIGNING_ISSUER_ID:?WPCT_MACOS_SIGNING_ISSUER_ID is required. Configure the macOS signing credentials in Buildkite.}"
+if ! printenv WPCT_MACOS_SIGNING_PRIVATE_KEY | grep -q .; then
+	echo "WPCT_MACOS_SIGNING_PRIVATE_KEY is required. Configure the macOS signing credentials in Buildkite." >&2
+	return 1
+fi
```
AliSoftware, line 12: Based on my tests if the variable is not set (or null) the error message printed by bash is already prefixed by the name of the undefined variable:
```bash
localhost:~# echo ${FOO:?missing env var in CI secrets}
bash: FOO: missing env var in CI secrets
```
So I think the message can be kept even simpler and doesn't need to re-state the variable name.

As for the 3rd variable, we already handle the case of it missing just a couple of lines later (L31-L34), so checking it early only a couple of lines before seems redundant?

Given all this I suggest this simpler version:

```suggestion
# Note: The Buildkite command enables `set -eu` before sourcing this script.
export APPLE_API_KEY_ID="${WPCT_MACOS_SIGNING_KEY_ID:?missing env var in CI secrets}"
export APPLE_API_ISSUER="${WPCT_MACOS_SIGNING_ISSUER_ID:?missing env var in CI secrets}"
```

.buildkite/commands/setup_macos_code_signing.sh
```diff
@@ -2,17 +2,14 @@
 
 echo "~~~ :apple: Configure macOS code signing"
```
AliSoftware: I think we should better organize the log sections here and properly group items related to code signing together.

Because right now not only there are 2 sections about code signing interrupted by an install_gems section in the middle, but also the `---` from install_gems is a section of higher level than `~~~` but here they're mixed wrong.

I suggest reorganizing things like this instead:
```
echo "--- :ruby: Install gems"
install_gems

echo "--- :key: Configure macOS code signing"
… the 2 exports, mktemp, 3rd export, cleanup trap, writing of the key, and call to 'fastlane setup_code_signing'
```

.buildkite/commands/setup_macos_code_signing.sh
iangmaia: 👍 I shortened the key ID and issuer messages while retaining the Buildkite-specific guidance. I kept the early private-key check because the later write guard runs after gem installation and tmp dir creation, so it diagnoses a write failure instead of a missing credential. Updated on 493a6e46060302a3600a47636f4e65e6a7753b2f.

.buildkite/commands/setup_macos_code_signing.sh
iangmaia: 👍 Updated on 8540473edde619efe0463d53a9f582181b923aa0.

AliSoftware (approved):

PR #487, opened 2026-09-17
[Change] Publish the docs site per release, not on every trunk push

juanmaguitar:
## Why

The published guide at https://wordpress.github.io/contributor-toolkit/guide/getting-started.html is built from trunk on every push that touches `docs/`. Merging the Gutenberg stack therefore put five unreleased features in front of users: the latest release is v1.1.0, and nothing anyone can download has "Working on a GitHub issue", the Gutenberg site type, or the Gutenberg PR panel.

This is not a docs mistake. Docs shipping in the same PR as the change they describe is the rule here, and the consequence of that rule is that trunk's guide is always ahead of the newest build. What was wrong is publishing trunk.

## What changes

The site now tracks the latest **release**. Deploy runs only on `workflow_dispatch`, with a required `ref` input naming what to publish, so cutting a release advances the guide with one command:

```bash
gh workflow run docs.yml --ref trunk -f ref=v1.2.0
```

Two decisions a reviewer would otherwise have to reverse-engineer:

**Why the ref is an input rather than the dispatch target.** Dispatching on tag `v1.1.0` would run the copy of `docs.yml` that exists *at that tag*, whose deploy job is gated on `github.ref == 'refs/heads/trunk'`. It would build and silently skip the deploy. So the dispatch happens on trunk and the ref to publish travels as an input.

**Why the deploy guard still tests `github.ref`.** It means something narrower now, and the comment says so: it decides which copy of this workflow may hold the Pages OIDC token (the reviewed, merged one), not what gets published. `inputs.ref` decides that, and publishing an arbitrary ref is the point. The `github-pages` environment already has a deployment-branch policy allowing only `trunk`, so this is belt and braces.

Pushes to trunk and pull requests still build, they just never deploy: a dead link fails when it is proposed and again when it lands, rather than at release time.

Not in this PR: any change to `docs/` content. Trunk's guide stays as it is and goes live at v1.2.0.

## How to test this

Platform: any. This is CI-only, nothing in the app changes.

**Starting state:** this branch, and the repo's Actions tab.

1. Open this PR's **Docs** check. Expected: the `build site` job runs and passes; **no** `deploy to Pages` job appears. The old behaviour was the same for PRs, so this confirms nothing regressed.
2. In the run's summary page, look at the **Record what was built** step's output. Expected: a "Built `<sha>`" line naming the commit, and no "Not published" warning (that warning is for dispatches only).
3. After merge, go to Actions → Docs → **Run workflow**. Expected: a **"Git ref to publish"** box, required, with no prefilled value.
4. Run it with branch `trunk` and ref `v1.1.0`. Expected: build passes, `deploy to Pages` runs, the run summary names `v1.1.0` and commit `576d218`, and the live guide loses the Gutenberg pages: no "Working on a GitHub issue" in the sidebar, and `/guide/gutenberg-issues.html` 404s.
5. The sharp-edge path: run it with branch `v1.1.0` (the tag) instead of trunk. Expected: it builds, the deploy is skipped, and the run summary carries a "⚠️ Not published" line telling you to re-run with `--ref trunk -f ref=<tag>`. Without that line this would be an all-green run that published nothing.

**What must not have happened:**

- A deploy triggered by a plain push to trunk. The whole point is that merging docs no longer publishes them.
- The site published from trunk while the input said `v1.1.0` — check the deployed page, not just the green check.
- A release publish cancelled mid-flight by an unrelated docs merge. Workflow-level concurrency cancels jobs too, so dispatches get their own concurrency group; the job-level `docs-deploy` group does not cover this.

## Risks and limitations

3 [fix here] · 2 [follow-up] from the review pass, all 5 addressed. Detail in the collapsed block.

- The guide is now stale by design between releases. Anyone reading the live site while trunk is ahead sees the last release, which is the intent, but it does mean a docs fix for a *released* bug also waits for the next release unless someone dispatches deliberately (which they can, at any time, with any tag).
- Whoever dispatches decides what goes live. `inputs.ref` can name an unmerged branch or a PR merge ref, so "the site is reviewed code" is now a habit rather than a branch rule. The build job holds nothing worth stealing (`contents: read`, `persist-credentials: false`, and the deploy job checks nothing out), so this is not a token-escalation path. Called out in the file's own comment.
- Steps 3-5 above cannot be exercised before merge: `workflow_dispatch` only offers the inputs declared in the copy of the file on the default branch. What *was* verified locally is that the v1.1.0 docs still build clean under today's toolchain (`npm ci --prefix docs` + `vitepress build` at tag `v1.1.0`, no dead links, 19 guide pages, no `gutenberg-issues.html`), so the restore dispatch is not going to fail on content.

## Related

Follow-up to the Gutenberg docs stack: #486, #481, #480, #476, #471.

---

<details>

**Reverting `docs/` on trunk to its v1.1.0 state.** Rejected: it fights the repo rule that docs ship with the change that makes them true, and every reverted page would have to be un-reverted at v1.2.0, with the Gutenberg screenshots going in and out of the tree twice.

**A long-lived `docs-live` branch that auto-deploys, fast-forwarded at each release.** Rejected as one more branch to maintain for the same outcome a dispatch already gives.

**A default value for the `ref` input.** Rejected: a hardcoded default names one release tag and is wrong from the next release onwards. Resolving "latest release tag" in a step was the alternative, and it adds a `gh` call and a failure mode to a job that runs twice a year. Typing the tag also makes the run record say what went live.

</details>

<details>

3 [fix here] · 2 [follow-up] — all 5 addressed in `628820c`.

- 🟡 security, `docs.yml` summary step: the ref was interpolated into `run:` through `${{ }}`, which substitutes into the script's source before bash parses it, so a ref name carrying `$( )` or a quote would have been code. Now passed through `env:` and read as `"$REF"`, the way `download-stats.yml` passes its values, with `set -euo pipefail` added. Verified by running the step locally with `REF='v1.1.0$(echo PWNED)"'`: printed literally.
- 🟡 architecture, concurrency: `docs-build-${{ github.ref }}` put a release dispatch and a routine trunk push in the same group, and workflow-level concurrency cancels the whole run, jobs included, so a docs merge could cancel a publish mid-flight. Dispatches are now keyed on `github.run_id`.
- 🔵 architecture, checkout: `ref: ${{ inputs.ref || github.ref }}` was not the no-op it claimed for push events. Checkout's own default is `github.sha`; `github.ref` would build whatever the branch tip is when the step runs. Changed to `github.sha`.
- 🔵 [follow-up] security: the header paragraph this diff removed was the one saying the deploy never runs on PR code. That property genuinely changes, so it is now stated on the deploy guard rather than dropped.
- 🔵 [follow-up] architecture: a dispatch on a non-trunk ref was an all-green run that published nothing. The summary step now writes a "Not published" warning naming the correct invocation.

The pass also confirmed, case by case, that the new guard holds: `pull_request` (fork included) fails the event test; a fork cannot dispatch at all; a branch or tag dispatch fails the ref test.

- **Review:** completed — `/self-review` per AGENTS.md, judgement pass in a fresh subagent context given only the diff and `code-review.instructions.md`; reviewed head = working tree over base `c91908c` (uncommitted at review time; the reviewed content is now committed as `628820c` with the five findings applied on top). Deterministic layer: `npm run lint` clean, `npm test` 1547 pass / 0 fail / 2 skipped. `actionlint` is not installed here and Docker was not running, so the workflow was checked by YAML parse and by executing the new shell step locally against both event shapes.
- **Since review:** `c91908c` → `628820c` — the five findings, plus one CONTRIBUTING.md wording fix. No re-review; the changes are the review's own remedies.

</details>

<details>
<summary>Implementation notes</summary>

`inputs` is a recognised context here because the workflow declares `on.workflow_dispatch.inputs`; on `push` and `pull_request` it evaluates to empty rather than failing expression evaluation, so `inputs.ref || github.sha` falls through cleanly.

`CONTRIBUTING.md` changes in the same PR because its "The documentation site" section stated the old behaviour in prose, and this change makes that sentence false.

</details>


juanmaguitar:
@coderabbitai review

juanmaguitar:
Both findings looked at.

**SC2016 (minor)** — fixed in 043883e: directive added, with a line saying why the quoting is deliberate.

**Authorization bypass (major)** — not fixed, and I think the severity is wrong rather than the observation. `workflow_dispatch` requires write access, so "Reachability: External" does not hold. More to the point, the trunk ruleset has `required_approving_review_count: 0`, so before this PR anyone with write access could publish arbitrary content to the live site by merging their own PR and letting the push trigger fire. The bar to publish was write access then and is write access now; what changed is that publishing is deliberate instead of a side effect of merging, and that fewer events reach the deploy job at all.

What the finding does correctly identify is that `inputs.ref` can name an unmerged branch or a PR merge ref, so "the site carries reviewed code" is now the dispatcher's habit rather than a branch rule. That is already stated in the PR's Risks section and in the workflow's own comment on the deploy guard. Gating `inputs.ref` on `refs/tags/v*` would close it and costs nothing at release time, at the price of not being able to publish a docs hotfix without cutting a tag. Worth doing as a follow-up, not a merge blocker on a change that does not widen the existing boundary.

